Issue #22, opened 2023-11-21
Dockerize the zk playground MVP


tbraun96:
Before moving forward with turning mp-ecdsa and wt-frost protocols into gadgets, we need to ensure that any modifications internally do not break an existing integration test. We can dockerize the playground, then, run it as a github pipeline step.

PR #23, opened 2023-11-22
Job hooks and dockerization

### LINKED ISSUE #18: Add pre-job and post-job hooks as an abstraction to Jobs submitted to the job manager
Many job types we use have a format: perform some checks to see if we need to run a job, and if so, run the job. The final pattern is when the job finishes, we want some finalization logic (e.g., for the DKG, we would submit the public key on-chain).

Currently, we require that jobs satisfy `T: SendFuture<'static, ()>`. Under the new model, we would have a trait:

```rust
pub enum ProceedWithExecution {
    True,
    False
}

#[async_trait]
pub trait ExecutableJob: SendFuture<'static, ()> {
    async fn pre_job_hook(&mut self) -> Result<ProceedWithExecution, Error>;
    async fn post_job_hook(&mut self) -> Result<(), Error>;
}
```

Thus, jobs that are submitted must satisfy:

```rust
fn push_job<J: ExecutableJob>(&self, job: J)
```

tbraun96:
Closes #18 and #22 

Note: we are adding git LFS storage, but, it's only relevant for the pipeline to speed up the compilation process for the docker integration tests.


drewstone (approved):
LG if CI passes

PR #24, opened 2023-11-22
Streamline and refactor

tbraun96:
The purpose of this PR is to cleanup any rough edges inside the abstractions to allow making future gadgets easier and less repetitive.


drewstone (approved):
LG

PR #38, opened 2023-12-14
Implement ZkGadget with Runtime API

### LINKED ISSUE #13: Implement WebbGadgetModule for ZkGadget (use the Jobs pallet runtime API)
Under the WebbGadgetModule implementation, we have a `get_next_job` function. We need to properly use the client API to query the blockchain, getting the proper `n` value, before creating the job.

shekohex:
Closes #13


zk-gadget/src/mock.rs
```diff
@@ -0,0 +1,399 @@
...
+  }
+}
+
+pub fn to_account_id32(id: u8) -> AccountId32 {
+    AccountId32::new([id; 32])
+}
+
+pub fn alloc_available_ip_addr() -> SocketAddr {
+    // use port 0 to let the OS allocate a port
+    SocketAddr::new([127, 0, 0, 1].into(), 0)
```
tbraun96, line 321: Are we sure using `0` will automatically allocate a new port? In my experience, the allocation doesn't happen until we call `bind` on a port. So, if a client uses this socket address, it won't be able to connect since 127.0.0.1:0 is not an actual port.

zk-gadget/src/mock.rs
```diff
@@ -0,0 +1,399 @@
...
+            .lock()
+            .execute_with(|| System::finalize());
+        let summary = FinalizeSummary::<Block> {
+            finalized: vec![header.hash()],
+            header,
+            stale_heads: vec![],
+        };
+        let notification = FinalityNotification::from_summary(summary, tx.clone());
+        Some(notification)
+    }
```
tbraun96, line 246: Note that the underlying AbstractGadget implementation will keep polling this in a loop since it's not actually awaiting anything. It will keep sending results in probably too fast.

zk-gadget/src/tests.rs
```diff
@@ -0,0 +1,9 @@
+use crate::mock::*;
+
+#[test]
+fn gadget_starts() {
+    let (ext, runtime) = new_test_ext();
+    ext.lock().execute_with(|| {
```
tbraun96, line 6: Note that `lock()` is used elsewhere in the code *while* this lock for the test is running. Since execute_with requires `&mut self`, an `RwLock` wouldn't work either. I don't think this pattern will work. We will need to think of something else.

PR #51, opened 2024-01-15
feat: add CGGMP21 Protocol to the gadgets built on top of `dnfs-cggmp21`

### LINKED ISSUE #50: [TASK] Add `dnfs/cggmp21` Gadget
- [ ] Add support for CGGMP21 gadget based on https://github.com/dfns/cggmp21
- [ ] Add tests for CGGMP21 gadget

shekohex:
Closes #50


test-utils/src/mock.rs
```diff
@@ -143,12 +143,14 @@ impl RolesHandler<AccountId> for MockRolesHandler {
...
-        }
+        let validators = (0..8).map(id_to_pair).collect::<Vec<_>>();
+        validators.iter().find_map(|p| {
+            if p.public() == address {
+                Some(p.public().to_raw_vec())
+            } else {
+                None
+            }
+        })
     }
```
shekohex, line 154: That was the bug where we couldn't submit jobs

shekohex:
@tbraun96 I've noticed that after adding the on-chain verification of DKG, the mp-ecdsa gadget fails in both Keygen and Signing.
The Keygen was failing due to we did not sign the public key before we submit our result on-chain; I've fixed that part, and it is working.

However, the signing does not work, it errors on chain with a [`SigningKeyMismatch`](https://github.com/webb-tools/tangle/blob/b3b9eb8d863f965fe3f59071d36ef235655dda1d/pallets/dkg/src/functions.rs#L243-L258) Error; which means the recovered key from the signature does not match the one we submitted in phase one.

Please note that these issues are currently in the mp-ecdsa gadget, the new gadget implemented in this PR works fine for both Keygen and signing.

drewstone (changes requested):

drewstone (changes requested):
Small fixes from a past review that never submitted

PR #52, opened 2024-01-17
Threshold BLS integration

drewstone:
Currently branching off of #51


protocols/threshold-bls/src/state_machine/keygen.rs
```diff
@@ -0,0 +1,216 @@
...
+        .map_err(|e| {
+            Error::IoError(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                e.to_string(),
+            ))
+        })?;
+    tracer.msg_sent();
+
+    // Round 2
+    tracer.round_begins();
```
tbraun96, line 189: This code looks like blocking code. We will need to wrap all blocking code inside a `tokio::task::spawn_blocking` closure to ensure high performance. Basically every `tracer.X` function should be wrapped.

drewstone: So this is the type of logic I found inside the cggmp's implementation of keygen. Can you verify all that's needed here @tbraun96. At some point I want you to take this over. Mostly trying to establish here some additional flows for non-Zengo MPC libraries, since it seems like we may want to easily hook anything up to round-based.

- https://github.com/dfns/cggmp21/blob/m/cggmp21/src/keygen/threshold.rs#L120

PR #54, opened 2024-01-22
Draft: Add macros

tbraun96:


protocols/stub/src/protocol.rs
```diff
@@ -0,0 +1,90 @@
...
+        &self.account_id
+    }
+
+    fn role_type(&self) -> RoleType {
+        RoleType::LightClientRelaying
+    }
+
+    fn is_phase_one(&self) -> bool {
+        true
+    }
```
shekohex, line 61: Could you please explain what do these methods do? As an implementer of a gadget, what should I return or expect?

shekohex (changes requested):

protocols/stub/src/protocol.rs
tbraun96: Added relevant documentation

drewstone (approved):
Looks good, the docs and such will need to be updated once other PRs are updated with these changes given my PR has naming convention updates. So @tbraun96 @shekohex as we get the other PRs updated don't forget to update these docs so naming is consistent.

PR #67, opened 2024-02-08
Add FullProtocolConfig trait + Automatic Test Generation

tbraun96:


test-utils/src/lib.rs
```diff
@@ -45,3 +50,139 @@ pub fn remove_job(role_type: RoleType, job_id: JobId) {
...
+                    .execute_with_async(|| {
+                        assert_eq!(1, 1);
+                    })
+                    .await
+            }
+
+            #[tokio::test(flavor = "multi_thread")]
+            async fn test_externalities_keygen() {
+                test_utils::setup_log();
+                const N: usize = 3;
```
drewstone, line 93: These values should become part of the macro. We need to be able to test and even consider benchmarking under diff settings.

drewstone (changes requested):

drewstone:
Please link this README in the main readme too @tbraun96. And keep asking yourself if this is sufficient for implementing new protocols and if it's clearly linked from the proper places.

PR #76, opened 2024-02-20
Update Gossip Networking

shekohex:
Needs: https://github.com/webb-tools/webb-rs/pull/184
Related: https://github.com/webb-tools/tangle/pull/489

### DIFF
--- a/gadget-common/src/gadget/network/gossip.rs
+++ b/gadget-common/src/gadget/network/gossip.rs
@@ -1,30 +1,21 @@
-//! A DKG Gossip Engine that uses [`sc_network::NetworkService`] as a backend.
+//! A Gossip Engine that uses [`sc_network::NetworkService`] as a backend.
 //!
 //! In a nutshell, it works as follows:
 //!
-//! 1. You create a new [`NetworkGossipEngineBuilder`] which does not require any setup for now,
+//! 1. You create a new [`NetworkGossipEngineBuilder`] which does require the protocol name and
+//!    keystore,
 //! 2. You call [`NetworkGossipEngineBuilder::build`] to get two things:
 //!   - a [`GossipHandler`] that is a simple background task that should run indefinitely, and
 //!   - a [`GossipHandlerController`] that can be used to control that background task.
 //!
 //! The background task ([`GossipHandler`]) role is to listen, and gossip (if enabled) all the
-//! DKG messages.
+//! messages.
 //!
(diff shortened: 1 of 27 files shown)


tbraun96:
> > We should consider editing the mock.rs file's `new_test_ext` function to use the gossip engine here instead of the in-memory network as part of this PR.
> 
> This would mean we would set up an actual network, no? Currently tests are running everything in a single binary.

Yes. It would still work though, since everybody would be bound to localhost.

gadget-common/src/gadget/network/gossip.rs
```diff
@@ -592,14 +576,36 @@ impl<B: Block + 'static, KBE: KeystoreBackend> GossipHandler<B, KBE> {
...
+                    .error(format!("Failed to decode message payload: {e:?}"));
+                return;
+            }
+        };
+
+        if !self
+            .keystore
+            .pair()
+            .public()
+            .verify(&payload_and_sig.payload, &payload_and_sig.signature)
```
drewstone, line 601: Does this hash the payload under the hood? Is the signature over the hashed payload? We should document this and the functions a lot more please.

drewstone (changes requested):

gadget-common/src/gadget/network/gossip.rs
shekohex: We should do a simple ecdsa verify https://docs.rs/sp-io/latest/sp_io/crypto/fn.ecdsa_verify_prehashed.html and hash the payload using https://docs.rs/sp-io/latest/sp_io/hashing/fn.keccak_256.html. Same when we do signing.
Your implementation here will fail since it will try to verify that this message signed by us! Which is not the case, since this message is received from the network. We have the `who`'s peer id, which is mapped to their account id (public key).

shekohex: I will refactor this part @tbraun96 dw

tbraun96: > Your implementation here will fail since it will try to verify that this message signed by us! Which is not the case, since this message is received from the network. We have the who's peer id, which is mapped to their account id (public key).

Correct, my mistake. In this case, get their public key and verify the signature. There's no need for hashing as both the `verify` and the `sign` use blake2b under the hood before actually signing/verifying the data. No need for double hashing with keccak.

drewstone (approved):

PR #82, opened 2024-02-23
Signing and Verification + Fix Prometheus Config Issue

tbraun96:


gadget-common/src/full_protocol.rs
```diff
@@ -10,11 +10,15 @@ use crate::prometheus::PrometheusConfig;
 use crate::protocol::AsyncProtocol;
 use crate::Error;
 use async_trait::async_trait;
+use frame_support::__private::sp_io::crypto::ecdsa_verify_prehashed;
```
drewstone, line 13: This is a bad import, you should use `sp_io` directly.

drewstone (changes requested):

PR #96, opened 2024-03-07
Secp256r1 and Stark for DFNS

tbraun96:


drewstone:
Cherry picked this commit into my branch.

PR #97, opened 2024-03-08
feat: MVP Gadget Shell

### LINKED ISSUE #92: [TASK] Create `gadget-libp2p`
## Minimum features needed

1. A p2p network (using libp2p, runs everywhere)
2. Authenticated.
3. Resilient to malicious nodes on the network.

### Messages specific requirements for Tangle.
5. Messages need to be prefixed w/ a `JobId` and a `RoundId` to prevent replayability of on-chain misbehavior reports.
6. Messages need to be signed by the ROLE key before they hit the wire UNLESS it's a non-restaker job participant.
7. Messages need to be verified by as being an on-chain ROLE key, UNLESS it's a non-restaker job participant.

### Handshake Process
On Connection, each peer needs to share a handshake message, the handshake message is just my `peer_id` + `signature` where the `signature` is the result of signing the `peer_id` using the role key.
Each peer verify will need to recover the signer (`role_key`) from the signed message and verify that they are a valid authority, their proxy account, or external participants (TBD)

In the event that a proxy key is being used for a restaker, there should be a method to quickly check the chain of `proxy key -> role key`

### Restakers Set

On each new session, we update our restakers set by query the on-chain data.

### LINKED ISSUE #93: [TASK] Add Gadget Shell with p2p network
- [ ] Update the gossip engine to use `gadget_libp2p` crate instead of `sc_network`. (@tbraun96 )

### LINKED ISSUE #94: [TASK] Implement keystore integration for local and remote execution

shekohex:
Adds the basic gadget shell binary, with the following features:
1. Loads keys from a keystore file (created by tangle or subxt)
2. Uses subxt to connect to tangle over RPC
3. Uses the RPC Client to query the chain's runtime API.
4. Adds libp2p Networking Implementation

Closes #92
Closes #93
Closes #94


shekohex:
CI does not have to be green, since it needs more work to be done by @tbraun96 too.

common/src/lib.rs
```diff
@@ -224,9 +224,10 @@ macro_rules! generate_setup_and_run_command {
         {
             use futures::TryStreamExt;
             let futures = futures::stream::FuturesUnordered::new();
+            let mut networks = std::collections::VecDeque::from(network);
 
             $(
-                let config = crate::$config::new(client.pop().expect("Not enough clients"), pallet_tx.clone(), network.pop().expect("Not enough networks"), logger.clone(), account_id.clone(), key_store.clone(), prometheus_config.clone()).await?;
+                let config = crate::$config::new(client.pop().expect("Not enough clients"), pallet_tx.clone(), networks.pop_front().expect("Not enough networks"), logger.clone(), account_id.clone(), key_store.clone(), prometheus_config.clone()).await?;
```
tbraun96, line 230: We pop front on the networks here to ensure consistent order, however, we do not do the same for the clients.

shekohex: It does not matter for the clients, they are the same.

PR #112, opened 2024-03-20
feat: watch for on-chain role changes

### LINKED ISSUE #95: [TASK] Implement chain query functionality for active runner roles

shekohex:
This PR adds a feature to automatically start and stop different protocols using the on-chain data based on the current restaker roles.

Closes #95

### DIFF
--- a/shell/src/shell.rs
+++ b/shell/src/shell.rs
@@ -143,7 +143,7 @@ where
 {
     let sub_account_id = subxt::utils::AccountId32(acco_key.public().0);
     // Create a loop that listens to new finality notifications,
-    // then it queries the chain for the current account roles
+    // queries the chain for the current restaking roles,
     // and then it starts the required protocols based on the roles.
     let mut current_roles = Vec::new();
     let mut running_protocols = HashMap::<HashedRoleTypeWrapper, tokio::task::AbortHandle>::new();
--- a/shell/src/shell.rs
+++ b/shell/src/shell.rs
@@ -144,7 +144,7 @@ where
     let sub_account_id = subxt::utils::AccountId32(acco_key.public().0);
     // Create a loop that listens to new finality notifications,
     // queries the chain for the current restaking roles,
-    // and then it starts the required protocols based on the roles.
+    // and then starts the required protocols based on the roles.
     let mut current_roles = Vec::new();
     let mut running_protocols = HashMap::<HashedRoleTypeWrapper, tokio::task::AbortHandle>::new();
 
--- a/shell/src/shell.rs
+++ b/shell/src/shell.rs
@@ -1,5 +1,6 @@
-use std::sync::Arc;
+use std::collections::{HashMap, HashSet};
 use std::time::Duration;
+use std::{hash::Hash, sync::Arc};
 
 use crate::{
     keystore::KeystoreContainer,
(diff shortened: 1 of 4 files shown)


shell/src/shell.rs
```diff
@@ -99,34 +85,121 @@ pub async fn run_forever(config: ShellConfig) -> color_eyre::Result<()> {
...
-        additional_params: (),
-        prometheus_config: PrometheusConfig::Disabled,
-        networks,
+    use RoleType::*;
+    use ThresholdSignatureRoleType::*;
+    let handle = match role {
+        Tss(DfnsCGGMP21Stark) | Tss(DfnsCGGMP21Secp256r1) | Tss(DfnsCGGMP21Secp256k1) => {
+            let node_input = NodeInput {
+                clients: vec![
+                    TangleRuntime::new(runtime.client()),
```
drewstone, line 106: Explain this more? Why do we need 4 runtimes?

shell/src/shell.rs
```diff
@@ -99,34 +85,121 @@ pub async fn run_forever(config: ShellConfig) -> color_eyre::Result<()> {
...
+    acco_key: sr25519::Pair,
+    logger: DebugLogger,
+    keystore: ECDSAKeyStore<KBE>,
+) -> color_eyre::Result<()>
+where
+    KBE: KeystoreBackend,
+{
+    let sub_account_id = subxt::utils::AccountId32(acco_key.public().0);
+    // create a loop that listen to new finality notifications
+    // then it queries the chain for the current account roles
```
drewstone, line 146:
```suggestion
    // queries the chain for the current restaking roles,
```

drewstone (changes requested):

shell/src/shell.rs
shekohex: It is not 4 different connections, they all share the underlying websocket connection here. (Same subxt client). For why different instances and not only one, because we have four different protocols that pull the finality notification stream at the same time and that may block.
So we create 4 different streams that are basically reading from a single stream (subxt connection)

shekohex:
Opened a few tasks for the missing systems:
#116

tbraun96 (commented):
Do we want to be able to start/stop gadgets on the fly depending on our roles? If so, this PR may need a redesign.

We could have a `HashMap<RoleType, tokio::sync::oneshot::Sender<()>>` for each role type's corresponding stopper_tx that stops running the spawned gadget. We just have to `select!` the gadget future and the stopper_rx in this case.

We would have a foreground task that periodically polls for our roles, and starts/stops gadgets as necessary.

On shell init, all we have to do is run this foreground task, and it will determine which gadgets to spawn and kill.

We could call this abstraction the `GadgetManager` or `ProtocolManager`.

shekohex:
That's exactly what these changes do. It will start a task to listen to finality notifications, on which we query the active roles for the current restaker, then we do a diff between the current and new roles to see what has changed (added, removed) then we start or stop services accordingly.

And there is no need for additional channels, for each task we already have an abort handle that will be used to abort the service if it has been removed from the roles.

drewstone:
We should also check there's no active jobs too just for extra safety.

tbraun96:
> That's exactly what these changes do. It will start a task to listen to finality notifications, on which we query the active roles for the current restaker, then we do a diff between the current and new roles to see what has changed (added, removed) then we start or stop services accordingly.
> 
> And there is no need for additional channels, for each task we already have an abort handle that will be used to abort the service if it has been removed from the roles.

In my experience, tokio's abort handle is not reliable. I have aborted tasks, yet they keep running. That's why I like using oneshot channels selected with the async function to make them cancellable.

tbraun96:
> We should also check there's no active jobs too for extra safety for each role in question.

Yes. On shutdown, we should also get the WorkManager and manually shutdown all the tasks.

shekohex:
What are the blocker for this PR?

PR #137, opened 2024-04-17
Shell Manager

tbraun96:


drewstone:
Still some clippy things here I'll leave for you to fix @tbraun96 is it ready otherwise?

PR #145, opened 2024-05-20
Refactor Shell Manager

tbraun96:


shell-manager/src/config.rs
```diff
@@ -0,0 +1,24 @@
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "Shell Manager",
+    about = "An MPC executor that connects to the Tangle network and runs protocols dynamically on the fly"
```
drewstone, line 7:
```suggestion
    about = "An program executor that connects to the Tangle network and runs protocols dynamically on the fly"
```

drewstone: Feel free to improve this too @tbraun96 .

drewstone (approved):

PR #154, opened 2024-05-21
[Phase 1]: Decouple event handler + client + transaction manager

tbraun96:


common/src/gadget/substrate/mod.rs
```diff
@@ -0,0 +1,299 @@
+use std::ops::Deref;
```
drewstone, line 1: I think everything in `gadget/substrate` should be moved out of common. It should be pulled into a directory called `gadget-impls` or something with subpackages that each implement connections to respective trait impls for this new API. Or at the very least, modularize everything into new packages. This is the "modular" in our modular restaking infrastructure that we need to keep driving home. CC @shekohex need to also ensure this matches our checklist module/breakdown for gadget-sdk.

common/src/gadget/substrate/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+use tangle_subxt::tangle_testnet_runtime::api::runtime_types::tangle_testnet_runtime::{MaxAdditionalParamsLen, MaxParticipants, MaxSubmissionLen};
+use gadget_core::gadget::manager::AbstractGadget;
+use crate::config::GadgetProtocol;
+use crate::gadget::{EventHandler, GeneralModule, MetricizedJob};
+use crate::prelude::{ClientWithApi, GadgetProtocolMessage, WorkManager};
+use crate::protocol::AsyncProtocol;
+use crate::tangle_runtime::AccountId32;
+
+/// Endow the general module with event-handler specific code tailored towards interacting with substrate
+pub struct SubstrateGadget<C, N, M> {
```
drewstone, line 17: Let's call this the Tangle gadget, immediately in the event handler you're using Tangle runtime to call into our job framework. It is much more specific than just Substrate.. it's Tangle. If we are to build a connection to Eigenlayer in the future for parsing AVS jobs it would be EigenlayerGadget not EthereumGadget.

drewstone (changes requested):
Looks good, some changes before we merge something like this or discussion we should have. I would like to get this merged and then quickly move onto v2 design. It's time we remove all v1 work from the gadget and push forward to the future system.

tbraun96:
@drewstone Yes, great ideas.

common/src/channels.rs
```diff
@@ -819,10 +826,10 @@ async fn wrap_message_and_forward_to_network<
     network: &N,
     user_id_mapping: &HashMap<UserID, ecdsa::Public>,
     my_user_id: UserID,
-    associated_block_id: <WorkManager as WorkManagerInterface>::Clock,
-    associated_session_id: <WorkManager as WorkManagerInterface>::SessionID,
-    associated_retry_id: <WorkManager as WorkManagerInterface>::RetryID,
-    associated_task_id: <WorkManager as WorkManagerInterface>::TaskID,
+    associated_block_id: <TangleWorkManager as WorkManagerInterface>::Clock,
```
drewstone, line 829: Why is Tangle used here? Can this be generic over any blockchain?

common/src/environments.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    fn build_protocol_message<Payload: Serialize>(
+        associated_block_id: <Self::JobManager as WorkManagerInterface>::Clock,
+        associated_session_id: <Self::JobManager as WorkManagerInterface>::SessionID,
+        associated_retry_id: <Self::JobManager as WorkManagerInterface>::RetryID,
+        associated_task_id: <Self::JobManager as WorkManagerInterface>::TaskID,
+        from: UserID,
+        to: Option<UserID>,
+        payload: &Payload,
+        from_account_id: Option<ecdsa::Public>,
```
drewstone, line 27: I think we need to generalize the account doing the key/signing of stuff too. Give that power to the AVS developer.

drewstone (changes requested):
Still some parts I'm unsure if we're generalizing as I intend.

common/src/channels.rs
drewstone: Yea, curious to sync on this. It does not seem correct semantically or in the right direction to have something `Tangle` named at this level. In fact there should be some config trait that captures most of these values and exposes functions to grab them like `retry`,`session`,`task`,`block_id`, or `timestamp` etc. that a task implementor can leverage and that don't all necessary need to be implemented for.

common/src/environments.rs
```diff
@@ -0,0 +1,65 @@
...
+        to_network_id: Option<ecdsa::Public>,
+    ) -> Self::ProtocolMessage;
+}
+
+#[derive(Default)]
+pub struct TangleEnvironment;
```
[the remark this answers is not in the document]
tbraun96: Re-organization and refactoring should happen once everything works in order to reduce the extra amount of entropy in involved with grasping the code. Though, yes, we could have our own crate per "Environment", or, hide these behind feature flags,

common/src/channels.rs
tbraun96: This is a draft, not finished.

tbraun96: We could, in its place (later down the road), encapsulate `retry,session,task,block_id, or timestamp` into some `EventMetadata` struct with a K,V hashmap (e.g., "retry -> int", "session" -> int ...). For now, my focus on doing one thing at a time, since decoupling our code is not an easy task

common/src/gadget/mod.rs
```diff
@@ -1,31 +1,35 @@
...
 
+pub mod core;
 pub mod message;
 pub mod network;
+pub mod tangle;
 pub mod work_manager;
 
 /// Used as a module to place inside the SubstrateGadget
-pub struct Module<C, N, M> {
+pub struct GeneralModule<N, M, Env: GadgetEnvironment> {
```
drewstone, line 27: I think `Module` is better for succinctness if you're just putting the word `General` on it. That doesn't describe it any more than keeping it as is.

common/src/gadget/work_manager.rs
```diff
@@ -1,11 +1,11 @@
 use crate::debug_logger::DebugLogger;
-use crate::gadget::message::{GadgetProtocolMessage, UserID};
+use crate::gadget::message::TangleProtocolMessage;
 use gadget_core::job_manager::WorkManagerInterface;
 use parking_lot::RwLock;
 use std::sync::Arc;
 
-pub struct WorkManager {
-    pub(crate) clock: Arc<RwLock<Option<<WorkManager as WorkManagerInterface>::Clock>>>,
+pub struct TangleWorkManager {
```
drewstone, line 7: We should sync here on this abstraction.

drewstone (approved):
LG to merge, lots more I'd like to make clearer. Good work nonetheless.

Issue #156, opened 2024-05-22
Gadget SDK - Logging Module


shekohex:

drewstone:
@Tjemmmic I've started this in #166, if you want to add to it that's the place.

PR #166, opened 2024-05-23
Add metrics and logging to sdk

drewstone:


shekohex:
We would actually need to use a different logging ecosystem than slong, in #156  you can see we are going to use tracing crate https://tracing.rs https://tokio.rs/#tk-lib-tracing. We can merge this Into my branch, but most of the work will be replaced with tracing.

PR #183, opened 2024-07-09
fix: build errors

shekohex:
Fixes build errors in eigen integration branch.


eigen-gadget/src/avs/mod.rs
```diff
@@ -36,7 +36,7 @@ sol!(
     #[derive(Debug)]
     #[sol(rpc)]
     Erc20Mock,
-    "contracts/out/Erc20Mock.sol/Erc20Mock.json"
+    "contracts/out/ERC20Mock.sol/ERC20Mock.json"
```
shekohex, line 39: File systems are case-sensitive, so we must use the exact same path, always.

PR #190, opened 2024-07-26
Use blueprint macros for the incredible squaring blueprint example

shekohex:
In This PR we started using the `job` macro for the incredible squaring blueprint example, the macro generates the metadata for the job and also generates the event handler, so that we are left with the following blueprint:

```rust
use gadget_sdk::job;
use std::convert::Infallible;

/// Returns x^2 saturating to [`u64::MAX`] if overflow occurs.
#[job(id = 0, params(x), result(_), verifier(evm = "IncredibleSquaringBlueprint"))]
pub fn xsquare(x: u64) -> Result<u64, Infallible> {
    Ok(x.saturating_pow(2u32))
}
```

And the following gadget:
```rust
use color_eyre::{eyre::OptionExt, Result};
use gadget_sdk::{
    events_watcher::tangle::TangleEventsWatcher, events_watcher::SubstrateEventWatcher,
    keystore::Backend, tangle_subxt::subxt,
};

use incredible_squaring_blueprint as blueprint;

#[tokio::main]
async fn main() -> Result<()> {
    color_eyre::install()?;
    let env_filter = tracing_subscriber::EnvFilter::from_default_env();
    let logger = tracing_subscriber::fmt()
        .compact()
        .with_target(true)
        .with_env_filter(env_filter);
    logger.init();

    let env = gadget_sdk::env::load()?;
    let keystore = env.keystore()?;
    let client = subxt::OnlineClient::from_url(&env.tangle_rpc_endpoint).await?;

    let sr25519_pubkey = keystore
        .iter_sr25519()
        .next()
        .ok_or_eyre("No sr25519 keys found in the keystore")?;
    let sr25519_secret = keystore
        .expose_sr25519_secret(&sr25519_pubkey)?
        .ok_or_eyre("No sr25519 secret found in the keystore")?;

    let mut seed = [0u8; 32];
    seed.copy_from_slice(&sr25519_secret.to_bytes()[0..32]);
    let signer = subxt_signer::sr25519::Keypair::from_secret_key(seed)?;

    let x_square = blueprint::XsquareEventHandler {
        service_id: env.service_id,
        signer,
    };

    tracing::info!("Starting the event watcher ...");

    SubstrateEventWatcher::run(
        &TangleEventsWatcher,
        client,
        // Add more handler here if we have more functions.
        vec![Box::new(x_square)],
    )
    .await?;

    Ok(())
}
```

The whole thing is under 60 LOC!


blueprints/incredible-squaring/deploy.ts
```diff
@@ -1,12 +1,12 @@
 import * as fs from "node:fs/promises";
```
drewstone, line 1: When it comes to the `gadget-cli`, we need to be automating this type of file as much as possible. Things like:
- Defining constants / variables to testnet/chain should be streamlined
- Deploying the contracts with the proper constructor arguments, our defaults can be empty (TODOs or other comments indicating what might be changed as if it's a coding exercise)
- Calling all jobs, since we are able to generate json, we should be able to iterate over all jobs and generate the TS scripts for calling things. This can likely be done in both TS and Rust. We want to streamline the devex / testing / deployment infra.

macros/blueprint-proc-macro-playground/src/lib.rs
```diff
@@ -20,12 +20,12 @@ impl std::fmt::Display for Error {
 
 impl std::error::Error for Error {}
 
-struct MyContext;
+pub struct MyContext;
```
drewstone, line 23: This is some start but I'm curious what your plans are here and if we are on the same page?

I expect the context to not be something created by the developer but rather something created by us at the SDK level. It should have a config that we can plug in things related to the Environment (or multiple Environments).

macros/blueprint-proc-macro/src/lib.rs
```diff
@@ -70,7 +80,14 @@ pub fn job(args: TokenStream, input: TokenStream) -> TokenStream {
...
 }
 
-/// TDOD: Documentation
+/// A procedural macro that annotates a function as a registration hook, mainly used
+/// for the metadata in the `blueprint.json`.
+/// # Example
+/// ```rust,no_run
+/// # use gadget_blueprint_proc_macro::registration_hook;
+/// #[registration_hook(evm = "MyRegistrationHook")]
+/// pub fn my_registration_hook(nft_id: u32);
```
drewstone, line 89:
```suggestion
/// pub fn my_registration_hook();
```

If we can't describe examples with respect to arguments in our docs, then we shouldn't be putting things like `nft_id` in the docs. It's confusing, I don't know what you intend to show me as a developer what I do with inputting an NFT ID. Where does this ID come from? If we don't have clarity don't add unclear details.

macros/blueprint-proc-macro/src/lib.rs
```diff
@@ -81,7 +98,14 @@ pub fn registration_hook(args: TokenStream, input: TokenStream) -> TokenStream {
...
 }
 
-/// TDOD: Documentation
+/// A procedural macro that annotates a function as a request hook, mainly used
+/// for the metadata in the `blueprint.json`.
+/// # Example
+/// ```rust,no_run
+/// # use gadget_blueprint_proc_macro::request_hook;
+/// #[request_hook(evm = "MyRequestHook")]
+/// pub fn my_request_hook(nft_id: u32);
```
drewstone, line 107:
```suggestion
/// pub fn my_request_hook();
```

Same here, please avoid superfluous details. Focus and clarity are really important here to ensure we're on the same page.

PR #206, opened 2024-08-06
Blueprint testing

tbraun96:


blueprint-manager/src/cli.rs
```diff
@@ -0,0 +1,40 @@
...
+use blueprint_manager::run_blueprint_manager;
+use blueprint_manager::utils;
+use gadget_common::gadget_io;
+use gadget_io::GadgetConfig;
+use shell_sdk::entry;
+use structopt::StructOpt;
+
+#[tokio::main]
+async fn main() -> color_eyre::Result<()> {
+    //color_eyre::install()?;
```
shekohex, line 11:
```suggestion
    color_eyre::install()?;
```

blueprint-manager/src/executor/event_handler.rs
```diff
@@ -1,141 +1,40 @@
...
-    // Handle registered events
+    let mut needs_update = false;
+
+    // Handle blueprint registered events
     for evt in registered_events {
         match evt {
             Ok(evt) => {
-                logger.info(format!("Registered event: {evt:?}"));
+                logger.info(format!("Blueprint registered event: {evt:?}"));
+                needs_update = true;
```
shekohex, line 37: we only update if the account who registered on that blueprint is equal to our account id.

blueprint-manager/src/executor/event_handler.rs
```diff
@@ -154,6 +53,8 @@ async fn handle_tangle_block(
                         "Removed all services for blueprint_id: {}",
                         evt.blueprint_id,
```
shekohex, line 54: why that's the case tho?

blueprint-test-utils/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        gadget_io::tokio::time::sleep(Duration::from_millis(100)).await;
+        // Most of what we need is here: api::tx().services() [..] .create_blueprint()
+        let call = api::storage().services().job_results(service_id, job_id);
+        let res = client.storage().at_latest().await?.fetch(&call).await?;
+        // client.tx() to send something to the chain
+
+        if let Some(ret) = res {
+            return Ok(ret);
+        }
+    }
```
shekohex, line 163: I would rather listen to events from the client and for each block I find the `JobResultSubmitted` event on which I will check for the call id and job_id.

blueprint-test-utils/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+pub async fn wait_for_completion_of_job(
+    client: &TestClient,
+    service_id: u64,
+    job_id: u64,
+) -> Result<JobResults, Box<dyn Error>> {
+    loop {
+        gadget_io::tokio::time::sleep(Duration::from_millis(100)).await;
+        // Most of what we need is here: api::tx().services() [..] .create_blueprint()
```
shekohex, line 155:
```suggestion

```

blueprint-test-utils/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        let call = api::storage().services().job_results(service_id, job_id);
+        let res = client.storage().at_latest().await?.fetch(&call).await?;
+        // client.tx() to send something to the chain
```
shekohex, line 158:
```suggestion

```

blueprint-test-utils/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        .sign_and_submit_then_watch_default(&call, user)
+        .await?;
+    let _res = res.wait_for_finalized_success().await?;
+    Ok(())
+}
+
+pub async fn wait_for_completion_of_job(
+    client: &TestClient,
+    service_id: u64,
+    job_id: u64,
```
shekohex, line 151:
```suggestion
    call_id: u64,
```

shekohex (changes requested):
Looks good overall, left a few comments

drewstone:
CI failing

blueprint-manager/src/executor/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        endpoint: gadget_config.url.clone(),
+    };
+
+    let tangle_environment = TangleEnvironment::new(subxt_config, acco_key, logger.clone());
+
+    let tangle_runtime = tangle_environment.setup_runtime().await?;
+    let runtime = ServicesClient::new(logger.clone(), tangle_runtime.client());
+    let mut active_shells = HashMap::new();
+
+    // With the basics setup, we must now implement the main logic
```
drewstone, line 152: Comments should be using `//`, if we can convert it that'd be great.

blueprint-manager/src/executor/event_handler.rs
```diff
@@ -1,141 +1,40 @@
...
     let registered_events = event.events.find::<Registered>();
     let unregistered_events = event.events.find::<Unregistered>();
     let service_initiated_events = event.events.find::<ServiceInitiated>();
     let job_called_events = event.events.find::<JobCalled>();
     let job_result_submitted_events = event.events.find::<JobResultSubmitted>();
 
-    // Handle registered events
+    let mut needs_update = false;
+
+    // Handle blueprint registered events
```
drewstone, line 32: Doesn't actually seem like these event handlers do anything yet?
- When we register for events we should download the blueprint binary right?
- When we unregister we should note that down (maybe remove binaries but also definitely not process any new services with those blueprints / side-effects to protect against).
- When we see a new service we should spawn the gadget?

Isn't this necessary to test?

drewstone (changes requested):

blueprint-manager/src/executor/event_handler.rs
tbraun96: The logic in the main loop of the work manager recognizes when an update happens, then, it delegates to the appropriate code path. So, yes, we do handle these events albeit, downstream.

drewstone (approved):

PR #216, opened 2024-08-13
feat(gadget-sdk): handle PreRegistration Event

### LINKED ISSUE #187: [TASK] Handle `PreRegistered` Event in the Shell
### Overview

The Shell should listen for `PreRegistered` events, this event includes a blueprint id and the operator account id, if the account id matches ours, we should download the blueprint and execute its gadget in the Registration Mode by setting `REGISTRATION_MODE_ON` env. The gadget will now start in the Registration mode and will try to register the current operator on that blueprint,  the required envs for this mode are the following:

1. `RPC_URL`: The remote RPC url for tangle
2. `KEYSTORE_URI`: the keystore file where the keys are stored and could be retrieved.
3. `DATA_DIR`: is an isolated path for where this gadget should store its data (database, secrets, …etc)
4. `BLUEPRINT_ID`: the active blueprint ID for this gadget
5. `REGISTRATION_MODE_ON` set to any value.

### References
See: https://github.com/webb-tools/tangle/issues/710

shekohex:
- [x] Handles `PreRegistration` on the gadget level.
- [x] Handles `PreRegistration` on the manager level (cc @tbraun96 )

Closes #187

### DIFF
--- a/blueprints/incredible-squaring/src/main.rs
+++ b/blueprints/incredible-squaring/src/main.rs
@@ -1,7 +1,11 @@
-use color_eyre::{eyre::OptionExt, Result};
+use color_eyre::{eyre::eyre, eyre::OptionExt, Result};
 use gadget_sdk::{
-    events_watcher::tangle::TangleEventsWatcher, events_watcher::SubstrateEventWatcher,
-    keystore::Backend, tangle_subxt::subxt,
+    events_watcher::{tangle::TangleEventsWatcher, SubstrateEventWatcher},
+    keystore::Backend,
+    tangle_subxt::tangle_testnet_runtime::api::{
+        self, runtime_types::sp_core::ecdsa, runtime_types::tangle_primitives::services,
+    },
+    tx,
 };
 
 use incredible_squaring_blueprint as blueprint;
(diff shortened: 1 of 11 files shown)


shell-manager/src/gadget/native.rs
```diff
@@ -117,17 +125,39 @@ async fn handle_github_source(
...
+                        // RPC_URL: The remote RPC url for tangle
+                        // KEYSTORE_URI: the keystore file where the keys are stored and could be retrieved.
+                        // DATA_DIR: is an isolated path for where this gadget should store its data (database, secrets, …etc)
+                        // BLUEPRINT_ID: the active blueprint ID for this gadget
+                        // REGISTRATION_MODE_ON set to any value.
+                        env_vars.push(("RPC_URL".to_string(), shell_config.url.to_string()));
+                        env_vars.push((
+                            "KEYSTORE_URI".to_string(),
+                            "file://./target/keystore".to_string(),
+                        ));
```
shekohex, line 149: That also should come from the CLI.

tbraun96: Would you prefer the shell/blueprint manager config, or, the individual shell config?

shekohex: It should be in the manager config nonetheless.

shekohex:
@drewstone this could be merged now.

drewstone (approved):
LG! We still need to create a trait or macro that exposes the custom registration function that will be executed offchain for constructing `RegistrationInputs` and then submitting it directly.

blueprints/incredible-squaring/src/main.rs
```diff
@@ -17,35 +21,46 @@ async fn main() -> Result<()> {
     logger.init();
 
     let env = gadget_sdk::env::load()?;
-    let keystore = env.keystore()?;
-    let client = subxt::OnlineClient::from_url(&env.tangle_rpc_endpoint).await?;
+    let client = env.client().await?;
+    let signer = env.first_signer()?;
+    if env.should_run_registration() {
```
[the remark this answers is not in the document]
shekohex: It is up to the blueprint developer to implement their custom registration code, this the whole idea, so no, this part cannot be abstracted, since each blueprint will require different registration params, if any at all.

PR #230, opened 2024-08-16
Add EVM event listener for Eigenlayer integration

drewstone:
- Adds event handler generation for eigenlayer contracts (using incredible squaring example)
- Add event handler errors
- Adds a job for eigen xsquare (still need to gossip aggregation within the job and make async, use network module)
- Separate tangle/eigen event handlers and parses them separately
- Refactors report file (reorganize ordering of impls)


drewstone:
@shekohex I'm a bit stuck on this if you're able to take a look. I do think we might want to avoid polling event listeners. They seem to consume RPCs much more than subscriptions. I don't think we should worry about whether a gadget has to sync to the latest state, perhaps its expected behavior that the gadgets are started BEFORE protocols begin execution or at the very least they begin operation at the tip of the chain their RPC is connected to.

shekohex:
> perhaps its expected behavior that the gadgets are started BEFORE protocols begin execution or at the very least they begin operation at the tip of the chain their RPC is connected to.

That's true, we actually do not sync anything, look at the substrate event watcher implementation, you will notice that we actually start from the current finalized block and continue from there every 6s. So we should do the same in this event listener, do not save any state and start from the current tip. we should infer the block time by getting the `current_block - 100` then read its timestamp, subtract the current block timestamp from the old block timestamp, and divide the result by 100. this would give us the avg block time.

The main reason we avoided subscription because it is state-full, However, I think we can use websockets now, since we do have the retry logic in place anyway.

macros/blueprint-proc-macro/src/eigenlayer/event_listener.rs
```diff
@@ -0,0 +1,77 @@
...
+    let callback = event_handler.callback().unwrap();
+
+    quote! {
+        use alloy_network::Network;
+        use alloy_provider::Provider;
+        use alloy_transport::Transport;
+        use alloy_sol_types::SolEvent;
+        use alloy_sol_types::SolInterface;
+        use std::ops::Deref;
+        use gadget_sdk::events_watcher::evm::Config;
```
shekohex, line 31: This will not work, you need to use the absolute path everywhere. The main reason behind this, when this code expands it will contain these use stmts, However if we have different jobs, they will also expand and there duplication of the same thing.

blueprints/incredible-squaring/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition.workspace = true
 license.workspace = true
 homepage.workspace = true
 repository.workspace = true
+spec_version = "0.1.0"
```
Tjemmmic, line 10:
```suggestion
spec_version.workspace = true
```

Cargo.toml
```diff
@@ -233,6 +233,8 @@ tracing-log = { version = "0.2", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 tracing-wasm = { version = "0.2.1" }
 tree_magic_mini = "3.1.5"
+
```
Tjemmmic, line 236:
```suggestion
spec_version = "0.1.0"
```

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -125,6 +129,9 @@ pub(crate) fn job_impl(args: &JobArgs, input: &ItemFn) -> syn::Result<TokenStrea
         #event_handler_gen
     };
 
+    // // Print the generated code
+    // println!("{}", gen);
+
```
Tjemmmic, line 134:
```suggestion

```

sdk/src/events_watcher/evm.rs
```diff
@@ -0,0 +1,250 @@
...
+pub trait EventHandler<T: Config>: Send + Sync {
+    /// The Type of contract this handler is for, Must be the same as the contract type in the
+    /// watcher.
+    type Contract: Deref<Target = alloy_contract::ContractInstance<T::T, T::P, T::N>>
+        + Send
+        + Sync
+        + 'static;
+
+    type Event: SolEvent + Clone + Send + Sync + 'static;
+    /// a method to be called with the event information,
```
Tjemmmic, line 46:
```suggestion
    /// A method to be called with the event information,
```

sdk/src/events_watcher/evm.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// A method to be called with the event information,
+    /// it is up to the handler to decide what to do with the event.
+    ///
+    /// If this method returned an error, the handler will be considered as failed and will
+    /// be retried again, depends on the retry strategy. if you do not care about the retry
+    /// strategy, use the [`EventHandler::handle_event`] method instead.
+    ///
+    /// If this method returns Ok(true), the event will be marked as handled.
+    ///
+    /// **Note**: this method is automatically implemented for all the event handlers.
```
Tjemmmic, line 75:
```suggestion
    /// **Note**: This method is automatically implemented for all the event handlers.
```

macros/blueprint-proc-macro/src/eigenlayer/event_listener.rs
drewstone: You're saying it won't work if these are imported already in the existing project, right?

PR #231, opened 2024-08-19
Make `{core, io, common}` no_std and WASM compatible

Serial-ATA:
`gadget-{core, io, common}` can now build in `#![no_std]` and `wasm32-unknown-unknown` environments.

## Notable changes

* Since this requires `Error` in `core`, I set the channel to **nightly**. This isn't long term, though. The feature should be out in a couple weeks.
	* To go with that, the `url` and `thiserror` crates use `no_std` compatible forks for now. Unfortunately, it'll probably be awhile before those get merged.
* `gadget-common` now builds with any combination of `--no-default-features`, and the `std`, `substrate`, and `wasm` features.
  * This is pretty messy, though. `common` will need to be reorganized eventually or this compatibility will be a pain to maintain in the future.
* `gadget_core::{job, gadget}` have been split into multiple modules to make it easier to gate their managers behind `std`.
* Removed some usage of `color_eyre`.
	* I haven't completely gotten rid of it yet, since it wasn't necessary to. We should be using `thiserror` instead, as `eyre` isn't meant for library errors.
	* For now, I've just been returning `String` errors in its place. I left TODOs to make concrete error variants later.

Outside of that, this is mostly just changing some dependency features and feature-gating.

Note that the commits won't work individually. I split them up per-crate to make it easier to review.


Serial-ATA:
The Clippy error is unrelated.

drewstone (approved):
LG, great work!

PR #233, opened 2024-08-20
Promote all dependencies to workspace

Serial-ATA:
All dependencies have been moved into the workspace manifest. I also updated them all to the latest version. `sysinfo` updated and started using `OsStr` instead of `str`, so I had to make some edits to `executor`.

This keeps all current dependencies. After this I'll look at what's actually needed.


shekohex (approved):
LGTM!

PR #237, opened 2024-08-21
Cleanup unused dependencies

Serial-ATA:
This removes unused dependencies and sorts the tables.

`common` depended on `color-eyre` just to re-export it. That re-export was only used by `tangle-environment`, so I just made a new `common::Error` variant for it to use instead (https://github.com/webb-tools/gadget/commit/0d8e38e5397f59af81e2d63ed6dadf56754dc685). Outside of that, there are no functionality changes.

### DIFF
--- a/gadget-io/Cargo.toml
+++ b/gadget-io/Cargo.toml
@@ -9,31 +9,31 @@ homepage.workspace = true
 
 [dependencies]
 cfg-if = { workspace = true }
-tracing = { workspace = true, default-features = false, features = ["log", "attributes"] }
 hex = { workspace = true }
-scale-info = { workspace = true, optional = true }
+multiaddr = { workspace = true, default-features = false }
 parity-scale-codec = { workspace = true, optional = true }
+sc-keystore = { workspace = true, optional = true }
+scale-info = { workspace = true, optional = true }
 serde = { workspace = true }
 sp-application-crypto = { workspace = true, default-features = false, features = ["full_crypto"] }
 sp-core = { workspace = true, features = ["serde", "full_crypto"] }
-multiaddr = { workspace = true, default-features = false }
-url = { workspace = true, default-features = false, features = ["alloc", "serde"] }
-thiserror = { workspace = true }
-tsify = { workspace = true, optional = true }
 sp-keystore = { workspace = true, optional = true }
-sc-keystore = { workspace = true, optional = true }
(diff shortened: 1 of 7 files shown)


gadget-io/Cargo.toml
```diff
@@ -9,52 +9,51 @@ homepage.workspace = true
...
+wasm-bindgen = { workspace = true, optional = true }
+wasm-bindgen-futures = { workspace = true, optional = true }
+wasmtimer = { workspace = true, optional = true }
 
 [features]
 default = ["std"]
 std = [
     "dep:tokio",
     "sp-application-crypto/std",
+    "dep:sp-keystore",
```
drewstone, line 44: Why do we need the `dep` prefix? The `sp` dependencies can compile without default features. Other ones that should compile without them are
- scale-info
- parity-scale-codec
- (maybe structopt)?

sdk/Cargo.toml
```diff
@@ -82,14 +75,15 @@ features = [
   "autonat"
 ]
 
-[target.'cfg(target_family = "wasm")'.dependencies.libp2p]
-workspace = true
-default-features = false
+# WASM-only deps
+[target.'cfg(all(any(target_arch = "wasm32", target_arch = "wasm64"), target_os = "unknown"))'.dependencies]
```
drewstone, line 79: Do we still need this block though?

drewstone: The SDK should just need `std` feature afaict?

drewstone (approved):
LG few comments

gadget-io/Cargo.toml
Serial-ATA: I just moved them in there since they don't get used outside of the standard keystore.

sdk/Cargo.toml
Serial-ATA: We'll still need a `wasm` feature. The `getrandom` js feature needs to be enabled for wasm32-unknown-unknown, but I can move that to the feature instead of having it automatically enabled for the target.

drewstone: I'm not entirely convinced we need getrandom, but I might be mistaken. My frame of reference is arkworks and substrate packages.

Serial-ATA: We don't actually use it directly, we're just required to explicitly enable the `js` feature to get it to build since `rand_core` is used in key generation.

PR #238, opened 2024-08-21
[feat] benchmark proc-macro

### LINKED ISSUE #234: [TASK] Create the benchmark proc-macro.

shekohex:
Closes #234


macros/blueprint-proc-macro/src/benchmark.rs
```diff
@@ -0,0 +1,79 @@
...
+    syn::custom_keyword!(threads);
+    syn::custom_keyword!(job_id);
+}
+
+/// `BenchmarkArgs` is a struct that holds the arguments for the `benchmark` macro.
+pub(crate) struct BenchmarkArgs {
+    /// The number of threads this benchmark should run with.
+    ///
+    /// `#[benchmark(threads = 4)]`
+    threads: syn::LitInt,
```
drewstone, line 17: I think there should be an option that tests your machine for how many threads it can reasonably support, then benchmarks with threads `1-num_threads`

drewstone: This way we can get metrics for a whole breadth of machine specs, no?

macros/blueprint-proc-macro-playground/src/lib.rs
```diff
@@ -104,6 +104,17 @@ fn report_service_health(uptime: f64, response_time: u64, error_rate: f64) -> Ve
     issues.concat()
 }
 
+// ==================
+//   Benchmarks
+// ==================
+#[benchmark(job_id = 0, threads = 2)]
+fn keygen_2_of_3() {
```
drewstone, line 111: The job was parameterized by (n,t), I think we should somehow extend the `params(n,t)` in a way that allows you to specify ranges for numbered parameters, thoughts?

shekohex: Yes we can do that, however the mechanism for doing this automatically is a bit complex (i.e matching the right t and n). A better approach would be a function generator that takes `i` as the run number and output the parameters for the job based on `i`.

shekohex:
Output:
```
Benchmark: keygen_2_of_3
Job ID: 0
Elapsed: 17.356583ms
vCPU: 2
RAM Usage: 5.23 MB
```

PR #240, opened 2024-08-23
Make a subset of gadget-sdk no_std compatible

Serial-ATA:
Each commit message speaks for itself, the only major change is the addition of `GadgetEnvironment::load_with_lock<R>()` where `R` is `lock_api::RawRwLock`. This allows callers to bring their own lock in `no_std` environments. `GadgetEnvironment::load()` is now `std` only, and will use `parking_lot::RawRwLock` as before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4369,7 +4369,6 @@ dependencies = [
  "parking_lot",
  "prometheus",
  "rand",
- "rand_core 0.6.4",
  "schnorrkel",
  "serde",
  "sp-core 34.0.0",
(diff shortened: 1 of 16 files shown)


drewstone:
Looks like CI wants improvements to the docs @Serial-ATA

Serial-ATA:
Fixed, I forgot to check Clippy after making those modules public.

drewstone:
Might need to update to latest Tangle deps to get the spelling error fix. I'm double-checking its in now.

drewstone (commented):
We should aim to get rid of the `wasm` feature and `getrandom`.

drewstone (approved):

PR #241, opened 2024-08-24
Move lints to workspace

Serial-ATA:


drewstone:
I'm finding pedantic lints a bit much TBH. I don't think we need to go overboard here unless you think it's totally worth it.

Serial-ATA:
Typically the move is to deny pedantic by default and just allow anything you don't agree with since there are so many nice lints in that family. But yeah, by default it is too much. There just weren't any `#![allow()]`ed lints yet, otherwise I would've moved them over too. Here's the config I've been using for all my libraries for example: https://github.com/Serial-ATA/lofty-rs/blob/0b903e9698f4b491c635a999ef7b50e0fd049ea1/Cargo.toml#L21-L56.

PR #244, opened 2024-08-26
Add more checks to CI

Serial-ATA:
* Renamed "Validate PR" to "CI" since it isn't run exclusively on PRs
* Added a `features` job to test that the crates build with each feature (including `--no-default-features`) using `cargo-hack`
* Added a test for WASM builds (commented out for now)

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -31,7 +31,6 @@ sp-core = { workspace = true, features = ["full_crypto"] }
 sp-io = { workspace = true, features = ["disable_allocator", "disable_panic_handler", "disable_oom"] }
 sp-runtime = { workspace = true, features = ["serde"] }
 substrate-prometheus-endpoint = { workspace = true, optional = true }
-subxt = { workspace = true, optional = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"], optional = true }
 tangle-subxt = { workspace = true, optional = true }
 
@@ -39,7 +38,7 @@ tangle-subxt = { workspace = true, optional = true }
 getrandom = { workspace = true, features = ["js"] }
 
 [features]
-default = ["substrate", "std", "tangle-testnet", "subxt/native"]
+default = ["substrate", "std", "tangle-testnet"]
 tangle-testnet = []
 tangle-mainnet = []
 substrate = [
@@ -47,9 +46,7 @@ substrate = [
     "dep:tangle-subxt",
     "dep:substrate-prometheus-endpoint",
     "dep:subxt-signer",
-    "dep:subxt",
     "gadget-core/substrate",
-    "subxt/jsonrpsee",
 ]
 std = [
     "dep:sqlx",
@@ -58,6 +55,7 @@ std = [
     "serde/std",
     "sp-io/std",
     "sp-runtime/std",
+    "tangle-subxt?/std",
 ]
-wasm = ["subxt-signer?/web", "subxt?/web", "subxt?/jsonrpsee"]
+wasm = ["subxt-signer?/web", "tangle-subxt?/web"]
 testing = []
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -38,7 +38,7 @@ tangle-subxt = { workspace = true, optional = true }
 getrandom = { workspace = true, features = ["js"] }
 
 [features]
-default = ["substrate", "std", "tangle-testnet"]
+default = ["substrate", "std", "tangle-testnet", "tangle-subxt/default"]
 tangle-testnet = []
 tangle-mainnet = []
 substrate = [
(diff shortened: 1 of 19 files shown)


sdk/src/lib.rs
```diff
@@ -30,8 +31,10 @@ pub mod env;
 pub mod store;
 
 /// Transaction Management Module
+#[cfg(any(feature = "std", feature = "wasm"))]
 pub mod tx;
 
+#[cfg(feature = "std")]
```
drewstone, line 37: This should be possible to be run in `no_std`? Therefore we should just have the `tangle_subxt/std` added to `std` feature in the TOML. If something needs to change on the package itself please handle it.

sdk/src/lib.rs
```diff
@@ -30,8 +31,10 @@ pub mod env;
 pub mod store;
 
 /// Transaction Management Module
+#[cfg(any(feature = "std", feature = "wasm"))]
```
drewstone, line 34: This is the type of thing I think feels wrong. Why if it can work in WASM/no_std do we need to specify either of these? Just add the relevant packages `std` feature to our SDK's `std` feature?

sdk/src/lib.rs
```diff
@@ -21,6 +21,7 @@ pub mod keystore;
 pub mod random;
 
 /// Blockchain Events Watcher Module
+#[cfg(any(feature = "std", feature = "wasm"))]
```
drewstone, line 24: Same here as the comment below. If it can work in `std` and `no_std` we shouldn't need to specify feature flags here. It should be configured in the TOML by adding the relevant packages to the `std` feature.

For example
````
std = ["package_that_works_in_no_std_and_std/std", ...]
```

drewstone (changes requested):

sdk/src/lib.rs
Serial-ATA: The `std` and `wasm` features have `dep:subxt`, which isn't no_std compatible.

drewstone: I believe parts of subxt are actually @Serial-ATA, can you double check. `subxt_core` should be: https://github.com/paritytech/subxt/issues/1479?

Serial-ATA: Yeah that works for `tangle_subxt`. I've been trying to convert that to have an `std` feature. In this case though, the only thing that's in `tx` relies on `OnlineClient` which is only available in `subxt`.

drewstone: This doesn't yet compile but I'm imagining in `tangle-subxt` something like
```
[dependencies]
parity-scale-codec = { workspace = true, default-features = false }
scale-info = { workspace = true, default-features = false }
subxt = { version = "0.37.0", default-features = false, optional = true }
subxt-signer = { version = "0.37.0", default-features = false, optional = true }
serde = { workspace = true, features = ["derive"], optional = true }

[features]
default = ["std"]
std = [
    "serde",
    "subxt/native",
    "subxt-signer/std"
]
web = [
    "subxt/web",
    "subxt-signer/web"
]
```

Serial-ATA: Yeah, that's pretty much what I have locally (still WIP). For the `tx` module though, there's nothing in there that can be used in `no_std` right now.

sdk/src/lib.rs
Serial-ATA: @drewstone That module has TODOs for more features anyway: https://github.com/webb-tools/gadget/blob/e910dc30a554a4a946328e9252166e8db5665a16/sdk/src/events_watcher/error.rs#L6

It'd be easier to handle that separation in its own PR, which I can do later.

shekohex (changes requested):

common/Cargo.toml
```diff
@@ -31,25 +31,22 @@ sp-core = { workspace = true, features = ["full_crypto"] }
...
-subxt = { workspace = true, optional = true }
 subxt-signer = { workspace = true, features = ["subxt", "sr25519"], optional = true }
 tangle-subxt = { workspace = true, optional = true }
 
 [target.'cfg(all(any(target_arch = "wasm32", target_arch = "wasm64"), target_os = "unknown"))'.dependencies]
 getrandom = { workspace = true, features = ["js"] }
 
 [features]
-default = ["substrate", "std", "tangle-testnet", "subxt/native"]
+default = ["substrate", "std", "tangle-testnet", "tangle-subxt/default"]
```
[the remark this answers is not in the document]
Serial-ATA: Yeah, that was just to get it to build while I focused on other things. I accidentally put `subxt/native` in the `default` feature instead of `std`: https://github.com/webb-tools/tangle/blob/d8cecdfe6a735977885b4a01c504b29823d9ba94/tangle-subxt/Cargo.toml#L24-L29

Fixing that would require `tangle-subxt` 0.3.0 (I think). I'll fix it soon.

shekohex (approved):

PR #246, opened 2024-08-26
[MEGA PR] Overhaul repo, add Eigenlayer AVS example, remove many crates, add testing, remove unused code

drewstone:


blueprints/incredible-squaring/Cargo.toml
```diff
@@ -40,5 +40,5 @@ alloy-transport = { workspace = true }
 blueprint-metadata = { workspace = true }
 
 [features]
-default = []
-gadget = ["dep:tokio", "dep:color-eyre", "dep:subxt-signer", "dep:tracing-subscriber"]
+#default = []
+default = ["dep:tokio", "dep:color-eyre", "dep:subxt-signer", "dep:tracing-subscriber"]
```
drewstone, line 44: I suggest using the following structure
```
default = ["std"]
std = ["dep:tokio", "dep:color-eyre", "dep:subxt-signer", "dep:tracing-subscriber"]
```

blueprints/incredible-squaring/src/main.rs
```diff
@@ -33,18 +36,21 @@ trait GadgetRunner {
...
 
-struct TangleGadgetRunner {
-    env: gadget_sdk::env::GadgetConfiguration,
+struct TangleGadgetRunner<R: lock_api::RawRwLock> {
+    env: gadget_sdk::env::GadgetConfiguration<R>,
 }
 
 #[async_trait::async_trait]
-impl GadgetRunner for TangleGadgetRunner {
+impl GadgetRunner for TangleGadgetRunner<dyn lock_api::RawRwLock<GuardMarker = GuardSend>> {
```
drewstone, line 44: Hmm this seems kinda messy, @Serial-ATA any suggestions to improve how this will integrate with our stack? I do want to hide some of these stuff as best we can when building blueprints. I think instead of `dyn ...` we should just add the trait to the `impl<L: lock_api::RawRwLock<...>>` ... TangleGadgetRunner<L>` if we need it like this @Tjemmmic

blueprints/incredible-squaring/src/main.rs
```diff
@@ -33,18 +36,21 @@ trait GadgetRunner {
...
     async fn register(&self) -> Result<()> {
-        let client = self.env.client().await?;
-        let signer = self.env.first_signer()?;
+        let client = self.env.client().await.map_err(|e| eyre!(e))?;
+        let signer = self.env.first_signer().map_err(|e| eyre!(e))?;
+
+        // TODO: Get ECDSA key from store as expected
+        let ecdsa_key = [0; 32];
```
drewstone, line 50: I believe the keystore / env has a method `first_ecdsa_signer()`?

blueprints/incredible-squaring/src/main.rs
tbraun96: We could put this RwLock as a type inside the `GadgetEnvironment` trait (which is about to be renamed, as this name is semantically misleading with respect to the new recent uses of an "environment"). We can put every type inside a single spot, allowing us to easily organize and change abstractions at varying levels of the stack.

blueprints/incredible-squaring/src/main.rs
```diff
@@ -23,50 +22,63 @@ use gadget_sdk::{
...
 use lock_api::{GuardSend, RwLock};
 use std::sync::Arc;
-use gadget_common::config::DebugLogger;
-use gadget_sdk::env::GadgetConfiguration;
-use gadget_sdk::network::gossip::GossipHandle;
+use alloy_signer::k256::PublicKey;
+use sp_core::Pair;
 
 #[async_trait::async_trait]
 trait GadgetRunner {
```
drewstone, line 42: Traits should now be in the SDK I think from @shekohex last PR. If you pull main in.

sdk/src/network/setup.rs
```diff
@@ -72,7 +72,7 @@ impl NetworkConfig {
     ) -> Self {
         Self {
             identity,
-            role_key,
+            ecdsa_key: role_key,
```
drewstone, line 75: Change the name of `role_key` above to `ecdsa_key` too.

drewstone (commented):
@Tjemmmic

drewstone:
Beginning or middle we should have solid update / review of this.

sdk/src/config.rs
```diff
@@ -84,21 +91,56 @@ pub struct GadgetConfiguration<RwLock: lock_api::RawRwLock> {
...
-
     /// The type of protocol the gadget is executing on.
     pub protocol: Protocol,
-
+    /// Bind port
+    pub bind_port: u16,
+    /// Bind addr
+    pub bind_addr: IpAddr,
+    /// logger
+    pub logger: Logger,
```
Serial-ATA, line 101: We should make sure that docs aren't just a repetition of the item name.

sdk/src/logger.rs
```diff
@@ -1,51 +1,84 @@
...
+
+    let _ = SubscriberBuilder::default()
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish()
+        .try_init();
+
+    std::panic::set_hook(Box::new(|info| {
+        log::error!(target: "gadget", "Panic occurred: {info:?}");
+    }));
+}
```
Serial-ATA, line 84:
```suggestion
```

This should be removed.

* `tracing_subscriber` should only ever be used by applications
* There isn't any benefit to setting a new panic hook here. It's actually a *loss* of information, since you no longer get the backtrace.

sdk/src/tx/tangle.rs
```diff
@@ -1,12 +1,15 @@
+use crate::logger::Logger;
+
 /// Send a transaction to the Tangle network.
 /// # Errors
 ///
 /// Returns a [`subxt::Error`] if the transaction fails.
-#[tracing::instrument(skip(client, signer, xt))]
+#[tracing::instrument(skip(client, signer, xt, logger))]
```
Serial-ATA, line 7:
```suggestion
#[tracing::instrument(skip_all)]
```

blueprint-test-utils/src/test_ext.rs
```diff
@@ -372,14 +446,14 @@ pub async fn new_test_ext<
 }
 
 pub struct LocalhostTestExt {
-    client: OnlineClient<SubstrateConfig>,
+    client: OnlineClient<TangleConfig>,
```
Serial-ATA, line 449:
```suggestion
    client: TangleClient,
```
The `TangleClient` and `TangleConfig` types should be used everywhere possible.

sdk/Cargo.toml
```diff
@@ -36,7 +43,8 @@ tokio = { workspace = true, features = ["rt-multi-thread", "parking_lot"], optio
 
 # Logging deps
 log = { workspace = true }
-tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter", "ansi", "tracing-log"] }
```
Serial-ATA, line 46:
```suggestion
```

sdk/src/config.rs
```diff
@@ -167,37 +267,57 @@ pub fn load(
...
+                ..
+            },
+        ..
+    } = additional_config;
+
+    let logger = logger.unwrap_or_default();
+
+    let data_dir = base_path;
+    let mut keystore_url = format!("{}", data_dir.display());
+    if !keystore_url.starts_with("file:/") && !keystore_url.starts_with("file://") {
```
Serial-ATA, line 301: The second condition is redundant

blueprints/incredible-squaring/src/eigenlayer.rs
```diff
@@ -0,0 +1,472 @@
...
+sol!(
+    #[allow(missing_docs)]
+    #[sol(rpc)]
+    IncredibleSquaringTaskManager,
+    "contracts/out/IncredibleSquaringTaskManager.sol/IncredibleSquaringTaskManager.json"
+);
+
+/// A wrapper around [`IncredibleSquaringTaskManagerInstance`] that can be dereferenced to its [`ContractInstance`].
+#[derive(Debug, Clone)]
+struct IncredibleSquaringInstance<T, P> {
```
drewstone, line 60: This seems like it can become auto-generated in the `job` macro no? Since we put these things directly in the macro already it should be.

blueprints/incredible-squaring/src/eigenlayer.rs
```diff
@@ -0,0 +1,472 @@
...
+            ContractInstance::new(
+                self.instance.address().clone(),
+                self.instance.provider().clone(),
+                Interface::new(JsonAbi::from_json_str(include_str!("../contracts/out/IncredibleSquaringTaskManager.sol/IncredibleSquaringTaskManager.json")).unwrap()),
+            )
+        })
+    }
+}
+
+impl<T, P> Deref for IncredibleSquaringInstance<T, P>
```
drewstone, line 91: Can we add this to SDK and macro system?

blueprints/incredible-squaring/src/eigenlayer.rs
```diff
@@ -0,0 +1,472 @@
...
+);
+
+/// A wrapper around [`IncredibleSquaringTaskManagerInstance`] that can be dereferenced to its [`ContractInstance`].
+#[derive(Debug, Clone)]
+struct IncredibleSquaringInstance<T, P> {
+    instance: IncredibleSquaringTaskManagerInstance<T, P>,
+    contract_instance: OnceLock<ContractInstance<T, P, Ethereum>>,
+}
+
+impl<T, P> IncredibleSquaringInstance<T, P>
```
drewstone, line 65: Can we add this to SDK and macro system?

blueprints/incredible-squaring/src/eigenlayer.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub fn set_operator(&mut self, operator: Operator<NodeConfig, OperatorInfoService>) {
+        self.operator = operator.into();
+    }
+}
+
+struct EigenlayerEventWatcher<T> {
+    _phantom: std::marker::PhantomData<T>,
+}
+
+impl<T: Config<N = Ethereum>> EventWatcher<T> for EigenlayerEventWatcher<T> {
```
drewstone, line 234: Can we add this to the macro system?

blueprints/incredible-squaring/src/eigenlayer.rs
```diff
@@ -0,0 +1,472 @@
...
+    let quorum_threshold_percentage = event.task.quorumThresholdPercentage;
+    (
+        number_to_be_squared,
+        task_created_block,
+        quorum_numbers,
+        quorum_threshold_percentage,
+    )
+}
+
+pub struct EigenlayerGadgetRunner<R: lock_api::RawRwLock> {
```
drewstone, line 211: I think this should be apart of the SDK. One idea would be to leverage a builder pattern or something that allows us to inject the custom AVS tooling into a generic runner so it can run without specifying all these boilerplate.

blueprints/tangle-avs-blueprint/src/main.rs
```diff
@@ -0,0 +1,130 @@
+pub mod runner;
```
drewstone, line 1: Tangle AVS blueprint is exclusively for Eigenlayer execution. It should not have a SubstrateEventListener IMO. It doesn't totally make sense to have a tangle operator run a tangle validator.. they'd just already be running a validator we wouldn't need a blueprint for it on Tangle, right?

cli/src/deploy.rs
```diff
@@ -299,6 +301,33 @@ fn bake_blueprint(
...
+
+    let (_, gadget) = blueprint_json["gadget"]
+        .as_object_mut()
+        .expect("Bad gadget value")
+        .iter_mut()
+        .next()
+        .expect("Should be at least one gadget");
+    let sources = gadget["sources"].as_array_mut().expect("Should be a list");
+
+    for (idx, source) in sources.iter_mut().enumerate() {
```
drewstone, line 313: Can we add comments?

cli/src/deploy.rs
```diff
@@ -299,6 +301,33 @@ fn bake_blueprint(
         convert_to_bytes_or_null(&mut job["metadata"]["name"]);
         convert_to_bytes_or_null(&mut job["metadata"]["description"]);
     }
+
+    let (_, gadget) = blueprint_json["gadget"]
```
drewstone, line 305: Can we add comments?

drewstone (commented):
More comments @Tjemmmic

blueprints/tangle-avs-blueprint/src/main.rs
Tjemmmic: I agree, I'm not 100% sure what exactly we want the Tangle AVS to do. I am removing it from this PR for now, since it is not going to be built out here anyway.

drewstone: For starters @Tjemmmic, we just want the AVS to run the validator client and register the keys involved that the validator will use. This should be sufficient to get on the testnet.

The secondary things that we will want to find process for are:
 - Registering for validator using tangle extrinsics (how will the AVS get a balance? can we do a cross-chain TX using hyperlane and gas subsidy?)
 - Implementing cross-chain messaging/eventually slashing logic (receive message from hyperlane bridge to the AVS contract).

These secondary things should come _after_ we get the first part done.

sdk/src/keystore/error.rs
```diff
@@ -1,4 +1,8 @@
 //! Keystore Errors
+#[cfg(not(any(feature = "std", feature = "wasm")))]
+use alloc::string::String;
+#[cfg(any(feature = "std", feature = "wasm"))]
+use std::string::String;
```
Serial-ATA, line 5: FYI, you can just use `alloc` for both `std` and `no_std` builds. But that's kind of a nit, I can fix that later.

sdk/src/config.rs
```diff
@@ -105,7 +104,7 @@ pub struct GadgetConfiguration<RwLock: lock_api::RawRwLock> {
     /// The Port of the Network that will be interacted with
     pub bind_port: u16,
     /// The Address of the Network that will be interacted with
-    pub bind_addr: IpAddr,
+    pub bind_addr: String,
```
Serial-ATA, line 107: Does this need to be a string? Using `core::net` types would be better if possible.

sdk/src/config.rs
```diff
@@ -372,6 +367,7 @@ impl<RwLock: lock_api::RawRwLock> GadgetConfiguration<RwLock> {
     /// * No sr25519 keypair is found in the keystore.
     /// * The keypair seed is invalid.
     #[doc(alias = "sr25519_signer")]
+    #[cfg(any(feature = "std", feature = "wasm"))]
```
Serial-ATA, line 370: Did something change in these keystore methods to make them no longer `no_std` compatible?

Tjemmmic: Yes, but I am restructuring to make it `no_std` compatible again. These fixes will be up soon.

PR #247, opened 2024-08-28
Move `gadget-io` into `gadget_sdk::io`

Serial-ATA:
This just moves `gadget-io` into `gadget_sdk::io`. Nothing has been removed or changed about it.

Most usage of `gadget-io` was for `gadget_io::tokio`, but it was usually in contexts where `tokio` was already a dependency. That wasn't the case for `core` and `common`, which now directly depend on `tokio`, but that doesn't really matter since they'll all get merged into `gadget-sdk` eventually anyway.


Serial-ATA:
Clippy errors are just from the typo in blueprint-manager and #243 needs to be merged.

shekohex:
> Clippy errors are just from the typo in blueprint-manager and #243 needs to be merged.

Merged, and rebased.

PR #274, opened 2024-09-16
[feat] Gadget Metadata

shekohex:
This PR introduces an easier way to define your gadget sources in your Cargo.toml

It is not, However, a new way that we created, it is actually encouraged by Cargo to let other tools embed extra metadata that could be later on used by these tools, see: []() for more information.

In our case, you can define your gadget metadata related to where we find your gadget binaries like the following:

```toml
[[package.metadata.gadget.Native.sources]]
owner = "webb-tools"
repo = "incredible-squaring"
tag = "0.1.0"
binaries = [
    { arch = "Amd64", os = "Linux", name = "amd64-linux-incredible-squaring-gadget" },
    { arch = "Arm64", os = "Linux", name = "arm64-linux-incredible-squaring-gadget" },
]
```

---

#### Open Questions
* In Github CI, we should update the `blueprint.json` with the SHA256 of the release binaries, that could be a separate task, or we could create our own github action. 
* Blueprint Specification (blueprint.json), potentially, should be attached to each github release.


drewstone:
What task is this for?

drewstone:
One more CI task to fix.

shekohex:
@drewstone it is failing since #246 I would mention @Serial-ATA here to investigate more in his PR.

Serial-ATA:
Yeah, #246 wasn't tested on the CI changes in #284. I'll take a look.

Issue #282, opened 2024-09-16
[CHECKLIST] Streamline Repo for DX Consistency


tbraun96:
### Tasks
After this epic is concluded, we will be able to confidently proceed forward with further, more challenging blueprints such as MPC with minimal additional complexity due to how the code currently works.

- [x] #277
- [x] #278
- [x] #279
- [x] #280
- [x] #281

PR #285, opened 2024-09-18
Streamline internal and external DX experience

tbraun96:
Addresses #282


drewstone:
Can you also fix up this broken CI task or remove it? It's pointless doesn't work in any branch.

blueprints/incredible-squaring/blueprint.json
```diff
@@ -57,9 +31,18 @@
   "request_hook": "None",
   "request_params": [],
   "gadget": {
-    "Wasm": {
-      "runtime": "Wasmtime",
-      "sources": []
+    "Native": {
```
tbraun96, line 34: We should consider adding `blueprint.json` to `.gitignore` since this value is 100% auto-generated.

sdk/src/keystore/mod.rs
```diff
@@ -210,43 +329,69 @@ pub trait Backend {
...
+/// that provide convenient access to keys
 pub trait BackendExt: Backend {
-    fn ecdsa_key(&self) -> Result<subxt_signer::ecdsa::Keypair, Error> {
+    #[cfg(any(feature = "std", feature = "wasm"))]
+    fn ecdsa_key(&self) -> Result<TanglePairSigner<sp_core_subxt::ecdsa::Pair>, Error> {
         let first_key = self
             .iter_ecdsa()
             .next()
-            .ok_or_else(|| Error::Ecdsa("No ECDSA keys found".into()))?;
+            .ok_or_else(|| str_to_std_error("No ECDSA keys found"))?;
```
Serial-ATA, line 340: Is there a reason to convert these all to `std::io::Error`?

sdk/src/logger.rs
```diff
@@ -68,3 +72,21 @@ impl Logger {
...
+}
+
+fn setup_log_no_panic_hook() {
+    let _ = SubscriberBuilder::default()
+        .without_time()
+        .with_span_events(FmtSpan::NONE)
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish()
+        .try_init();
+}
```
Serial-ATA, line 92: Note this: https://github.com/webb-tools/gadget/pull/246#discussion_r1759134224

Something like this shouldn't really be in here. It's a loss of information in that you're setting a new panic hook and making assumptions about the setup of the tracing subscriber.

sdk/src/config.rs
```diff
@@ -298,16 +299,17 @@ fn load_inner<RwLock: lock_api::RawRwLock>(
...
+                protocol,
                 ..
             },
         ..
-    } = additional_config;
+    } = config;
 
     let logger = logger.unwrap_or_default();
 
     let data_dir = base_path;
```
Serial-ATA, line 310: I'm still confused about the `data_dir_path`. Is it supposed to be the same as `keystore_url`? If I do `--base-path=":memory:"` then it won't be an actual *path*.

macros/blueprint-proc-macro-core/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub enum FieldType {
     Int64,
     /// A field of `u128` type.
     Uint128,
+    /// A field of `u256` type
+    U256,
```
Serial-ATA, line 31:
```suggestion
    Uint256,
```
Or is this something special that shouldn't follow the other variants?

sdk/src/config.rs
```diff
@@ -251,6 +255,8 @@ pub enum GadgetCLICoreSettings {
         blueprint_id: u64,
         #[structopt(long)]
         service_id: Option<u64>,
+        #[structopt(long, parse(try_from_str))]
+        protocol: Protocol,
```
Serial-ATA, line 259: Could just have it default to Tangle.

blueprints/incredible-squaring/blueprint.json
```diff
@@ -57,9 +31,18 @@
...
+      "sources": [
+        {
+          "fetcher": {
+            "Testing": {
+              "cargo_package": "incredible-squaring-blueprint",
+              "cargo_bin": "main",
+              "base_path": "/Users/nologik/commonwealth/gadget/blueprints/incredible-squaring"
```
Serial-ATA, line 41:
```suggestion
```

sdk/src/keystore/mod.rs
```diff
@@ -44,14 +44,133 @@ pub mod error;
...
 
-pub type TanglePairSigner = PairSigner<PolkadotConfig, subxt_core::ext::sp_core::sr25519::Pair>;
+#[cfg(any(feature = "std", feature = "wasm"))]
+#[derive(Clone, Debug)]
+pub struct TanglePairSigner<Pair> {
+    pub(crate) pair: subxt::tx::PairSigner<TangleConfig, Pair>,
+}
+
+#[cfg(any(feature = "std", feature = "wasm"))]
+pub type TangleSr25519PairSigner = TanglePairSigner<sp_core_subxt::sr25519::Pair>;
```
Serial-ATA, line 70: Theres a mix of `TangleSr25519PairSigner` and `TanglePairSigner<ecdsa::Pair>`. Why not make an alias for ECDSA as well? Or just use no aliases.

Serial-ATA (changes requested):

macros/blueprint-proc-macro-core/src/lib.rs
tbraun96: Special case

sdk/src/config.rs
tbraun96: I want these mandatory to not have silent bugs in the future that might cause the failing of passing the protocol argument.

blueprints/incredible-squaring/blueprint.json
tbraun96: I'm thinking we should just add blueprint.json to .gitignore since it is auto-generated.

sdk/src/keystore/mod.rs
tbraun96: We could, though, I'd have to make an alias for the others as well. Since `TangleSr25519PairSigner` is used much more than the related ecdsa pair, I make it a special case of having its own type.

sdk/src/logger.rs
drewstone: I think what @Serial-ATA is saying is still valid. Isn't getting a full backtrace going to be important for people other than yourself? If this is reducing the standard error information that is presented I think this is bad.

sdk/src/config.rs
drewstone: Please address this @tbraun96

sdk/src/keystore/mod.rs
```diff
@@ -44,14 +44,133 @@ pub mod error;
 /// Schnorrkel Support
 pub mod sr25519;
 
-use alloc::string::ToString;
+use crate::clients::tangle::runtime::TangleConfig;
+use crate::keystore::sp_core_subxt::crypto::{DeriveError, SecretStringError};
+use crate::keystore::sp_core_subxt::DeriveJunction;
+#[cfg(any(feature = "std", feature = "wasm"))]
+// TODO: Once subxt uses sp-core 34.0.0, we can simply use sp_core
+pub use crate::tangle_subxt::subxt::ext::sp_core as sp_core_subxt;
```
drewstone, line 52: @tbraun96 in the future please take initiative to make small changes that help us keep our repo clean too: https://github.com/paritytech/subxt/pull/1777

sdk/src/keystore/mod.rs
drewstone: If my PR gets through let's not use aliases. It's more and more unnecessary code that. Alex is right. We should be consistent with what we are going across the repo or we should not do something at all if its not needed.

sdk/src/keystore/mod.rs
```diff
@@ -44,14 +44,133 @@ pub mod error;
...
+
+#[cfg(any(feature = "std", feature = "wasm"))]
+pub type TangleSr25519PairSigner = TanglePairSigner<sp_core_subxt::sr25519::Pair>;
+
+#[cfg(any(feature = "std", feature = "wasm"))]
+impl<Pair: sp_core_subxt::Pair> sp_core_subxt::crypto::CryptoType for TanglePairSigner<Pair> {
+    type Pair = Pair;
+}
+
+#[cfg(any(feature = "std", feature = "wasm"))]
```
drewstone, line 77: Can someone remind me why we need to have this? If this works for any feature why can't it just not have a feature flag?

sdk/src/keystore/mod.rs
```diff
@@ -44,14 +44,133 @@ pub mod error;
...
+use subxt::ext::sp_core::Pair as PairSubxt;
+use subxt_core::tx::signer::{PairSigner, Signer};
+use subxt_core::utils::{AccountId32, MultiAddress, MultiSignature};
+#[cfg(any(feature = "std", feature = "wasm"))]
+use tangle_subxt::subxt;
 
-pub type TanglePairSigner = PairSigner<PolkadotConfig, subxt_core::ext::sp_core::sr25519::Pair>;
+#[cfg(any(feature = "std", feature = "wasm"))]
+#[derive(Clone, Debug)]
+pub struct TanglePairSigner<Pair> {
```
drewstone, line 65: Why do we need this at all? I'm confused, it's a wrapper over something that works and we're reimplementing all the traits just to call and prefix it with `Tangle`?

sdk/src/keystore/mod.rs
drewstone: We should consider specific errors here.

drewstone (changes requested):

sdk/src/keystore/mod.rs
tbraun96: It's a wrapper designed to be used throughout the codebase to minimize common issues that can arise when internal/external developers work with keys. To ensure consistency, it limits freedom by enforcing structure, which we lacked before.

sdk/src/config.rs
tbraun96: Yeah, there was certainly redundancy. I have merged it all into a single field, using uri's instead.

PR #311, opened 2024-09-24
feat(sdk): Remove Logger

### LINKED ISSUE #260: [TASK] Replace `Logger` with `tracing`
# Overview

`Logger` is simply a wrapper around `log` with a given ID: https://github.com/webb-tools/gadget/blob/main/common/src/debug_logger.rs

This can be replaced with `tracing`, which would allow us to provide that additional context, while also being able to:

* Stop manually calling `format!()`
* Stop having to pass a struct around everywhere we want logging

Serial-ATA:
closes #260


drewstone (approved):
Looks great!

Serial-ATA:
Actually, need to check a few tests first.

PR #317, opened 2024-09-25
Event listener

tbraun96:
Expect code to be moved around as I converge to the final, cleanest solution.


tbraun96:
Still need to generalize a bit more to increase flexibility, as well as potentially allow for multiple event listeners.

blueprint-manager/src/sdk/utils.rs
```diff
@@ -60,13 +60,6 @@ pub fn generate_process_arguments(
         format!("--bind-addr={}", gadget_config.bind_addr),
         format!("--bind-port={}", gadget_config.bind_port),
         format!("--url={}", gadget_config.url),
-        format!(
-            "--logger=Blueprint-{blueprint_id}-Service-{service_id}-{}",
-            opt.instance_id.clone().unwrap_or_else(|| format!(
-                "{}-{}",
-                gadget_config.bind_addr, gadget_config.bind_port
-            ))
-        ),
```
Serial-ATA, line 69: I forgot to update this. It's called `--log-id` now, btw.

tbraun96: Thanks!

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -17,6 +17,7 @@ mod kw {
     syn::custom_keyword!(verifier);
     syn::custom_keyword!(evm);
     syn::custom_keyword!(event_handler);
+    syn::custom_keyword!(event_listener);
     syn::custom_keyword!(protocol);
```
drewstone, line 21: Can you also remove this and remove the need for this? It seems superfluous given the introduction of this abstraction. Lmk if you think otherwise for now. I prefer the 2 birds w/ 1 stone approach whenever possible.

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -77,6 +78,35 @@ pub(crate) fn job_impl(args: &JobArgs, input: &ItemFn) -> syn::Result<TokenStrea
         }
     }
 
+    // TODO:
```
drewstone, line 81: TODO what?

sdk/src/event_listener/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+{
+    fn from(value: Watcher) -> Self {
+        Self {
+            listener: value,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+// TODO: Refactor up a level of abstraction after tests pass to truly use the EventListener interface
```
drewstone, line 53: I made a mistake calling it Eigenlayer protocol because I don't want you to associate EVM listeners with only Eigenlayer! They can be used on Tangle too. So we should just require you to specify an event listener and that will be the trigger for job execution. EVM listeners can also be instantiated in the `main.rs` to listen to contracts on any blockchain by setting up providers and contracts against any RPC anywhere. Let me know if this makes sense!

drewstone (changes requested):

macros/blueprint-proc-macro/src/job.rs
tbraun96: This is a good refactoring task. After much friction, I decided to shelve my changes and move forward since what we have now works, and will adds improvements to the job macro with respect to ordering params.

PR #322, opened 2024-09-30
Remove Git Dependencies for Publishing

Tjemmmic:
# Overview

Removes Git Dependencies so SDK can be published to Crates.io

### DIFF
--- a/sdk/src/keystore/backend/mem.rs
+++ b/sdk/src/keystore/backend/mem.rs
@@ -60,7 +60,7 @@ impl PartialOrd for BlsBn254PublicWrapper {
 
 impl Ord for BlsBn254PublicWrapper {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        self.0.to_bytes().cmp(&other.0.to_bytes())
+        self.0.g1().to_string().cmp(&other.0.g1().to_string())
     }
 }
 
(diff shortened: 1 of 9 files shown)


drewstone (approved):
Looks great! Awesome work getting this in quickly

sdk/src/keystore/backend/mem.rs
```diff
@@ -60,7 +60,7 @@ impl PartialOrd for BlsBn254PublicWrapper {
 
 impl Ord for BlsBn254PublicWrapper {
     fn cmp(&self, other: &Self) -> core::cmp::Ordering {
-        self.0.to_bytes().cmp(&other.0.to_bytes())
+        self.0.g1().to_string().cmp(&other.0.g1().to_string())
```
Serial-ATA, line 63: @Tjemmmic Does this even need to implement `Ord`?

Tjemmmic: It's required to use it as a Map's Key:

``` rust
fn bls_bn254_generate_new(&self, seed: Option<&[u8]>) -> Result<bn254::Public, Error> {
        let secret = bn254::generate_with_optional_seed(seed);
        let public = bn254::to_public(&secret);
        let old = self
            .bn254
            .write()
            .insert(BlsBn254PublicWrapper(public.clone()), secret); // <--- Here
        assert!(old.is_none(), "generated key already exists");
        Ok(public)
    }
```

``` rust
#[stable(feature = "rust1", since = "1.0.0")]
#[rustc_confusables("push", "put", "set")]
pub fn insert(&mut self, key: K, value: V) -> Option<V>
where
     K: Ord,
{
     match self.entry(key) {
         Occupied(mut entry) => Some(entry.insert(value)),
         Vacant(entry) => {
             entry.insert(value);
             None
         }
     }
 }
```

PR #324, opened 2024-09-30
Removes Remaining Git Dependencies

Tjemmmic:
# Summary

Removes the Git dependencies from `thiserror` and `url` that were leftover from `no-std` compatibility work.


shekohex (approved):
LGTM

PR #333, opened 2024-10-02
Event listener Docs + Wrapper Types

tbraun96:


docs/event_listeners.md
```diff
@@ -0,0 +1,235 @@
...
+    x: u64,
+    context: MyContext, // <-- The context type must be the first additional parameter
+    env: GadgetConfiguration<parking_lot::RawRwLock>,
+) -> Result<u64, Infallible> {
+    Ok(x.saturating_pow(2u32))
+}
+```
+
+## Wrapper Event Listeners
+Various wrappers exist to simplify how common operations are performed. For example, there is the `PeriodicEventListener`:
```
Serial-ATA, line 84:
```suggestion
Various wrappers exist to simplify how common operations are performed.
```

docs/event_listeners.md
```diff
@@ -0,0 +1,235 @@
...
+
+```rust
+/// Implement the [`EventListener`] trait for the Ticker struct. [`EventListener`] has two type parameters:
+/// 
+/// - T: is the type of the event that the listener listens for. This can be anything that is Send + Sync + 'static.
+/// In this case, we are using [`Instant`], which is a timestamp.
+///
+/// - Ctx: is the context type that the listener receives when constructed. This can be anything that is Send + Sync + 'static,
+/// with the special requirement that it is the *first* listed additional parameter in the [`job`] macro (i.e., a parameter not
+/// in params(...)). In this case, we are using [`MyContext`].
```
Serial-ATA, line 38: That seems like an odd restriction? Why not mark it with `#[context]` or check for a parameter named `context`?

docs/event_listeners.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Wrapper Event Listeners
+Various wrappers exist to simplify how common operations are performed. For example, there is the `PeriodicEventListener`:
+
+### PeriodicEventListener
+Some programs may only be interested in checking for events at regular intervals. In this case, the `PeriodicEventListener` can be used to simplify the process.
+
+A `PeriodicEventListener` is a wrapper that takes 4 type parameters:
+
+* `MSEC`: The number of milliseconds between each event check.
```
Serial-ATA, line 91: nit: The `Ticker` example above used `MS`, this one uses `MSEC`

docs/event_listeners.md
```diff
@@ -0,0 +1,235 @@
...
+* `T`: The inner event listener type
+* `Evt`: The event type
+* `Ctx`: The context type
+
+We can make a `PeriodicEventListener` that ticks every 5000ms to check the status of a web server
+
+```rust
+use reqwest;
```
Serial-ATA, line 99:
```suggestion
We can make a `PeriodicEventListener` that ticks every 5000ms to check the status of a web server using [reqwest](crates.io/crates/reqwest).

```rust
```

docs/event_listeners.md
tbraun96: I would also like to find an improvement over this. For now, this is just what was agreed upon by @shekohex and I.

shekohex: We could go with the more explicit path here with adding `#[context]` but we are aiming for less noise which would improve readability.

```rust
#[job(id = 1, params(x, y), result(_)]
fn my_job(context: Context, x: u64, y: u64) -> Result<u64, Inflatable> { ... } 
```
vs
```rust
#[job(id = 1, params(x, y), result(_)]
fn my_job(#[context] context: Context, x: u64, y: u64) -> Result<u64, Inflatable> { ... } 
```

sdk/src/event_listener/substrate.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[async_trait]
+pub trait SubstrateEventListenerContext<T: Config>: Clone {
+    fn rpc_url(&self) -> String {
+        "ws://127.0.0.1:9944".to_string()
+    }
+
+    async fn handle_event(
+        &self,
+        event: Block<T, OnlineClient<T>>,
```
shekohex, line 27: Is this a block or event? Please see https://docs.rs/subxt/latest/subxt/events/struct.Events.html instead of https://docs.rs/subxt/latest/subxt/blocks/struct.Block.html, unless you mean here a block, not an event.

shekohex (changes requested):

sdk/src/event_listener/substrate.rs
tbraun96: The `Block`, with respect to the `EventListener` trait, is a type that represents an event. The `Events` struct and `Block` structs belong to a different context outside the scope of this one.

docs/event_listeners.md
tbraun96: Adjusted the example above to use `MSEC` so it's all consistent.

docs/event_listeners.md
```diff
@@ -0,0 +1,235 @@
...
+Various wrappers exist to simplify how common operations are performed. For example, there is the `PeriodicEventListener`:
+
+### PeriodicEventListener
+Some programs may only be interested in checking for events at regular intervals. In this case, the `PeriodicEventListener` can be used to simplify the process.
+
+A `PeriodicEventListener` is a wrapper that takes 4 type parameters:
+
+* `MSEC`: The number of milliseconds between each event check.
+* `T`: The inner event listener type
+* `Evt`: The event type
```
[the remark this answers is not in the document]
Serial-ATA: @tbraun96 FYI this is just updated in the doc, should be changed on the struct too.

sdk/src/event_listener/substrate.rs
shekohex: It is confusing, the type is a Block, but you treat it as an event. If we are going all in blocks, then call it block not an event. Also, in your example, you did not show how to get the event? how to apply filters? ofc I know how, but you have to show them in the docs.

drewstone:
Points of feedback from meeting:
1. Event listener is required, move existing defaults out and into the SDK as options to choose from out of the box.
2. Aim for as little extra code as needed on Blueprint developer for specifying their event listeners. Less is more.
3. Docs for each section of event listeners: EVM Contract listeners, Tangle listener, Periodic Listeners, and ways to extend and build Custom Listeners on top of this.
4. Docs inside `tangle-docs`.

shekohex (changes requested):
There is still more work to be done here for sure, but I believe we are deviating from the original goal of this PR which is to support other even watchers other than the default ones (Substrate, EVM), and other types of Event Handlers, so why not adopting what we currently have and add more events watcher, like the common ones we see would be good for other blueprints?

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -87,30 +92,79 @@ pub(crate) fn job_impl(args: &JobArgs, input: &ItemFn) -> syn::Result<TokenStrea
...
+                            let ctx = (ctx.client.clone(), std::sync::Arc::new(ctx.clone()) as gadget_sdk::events_watcher::substrate::EventHandlerFor<gadget_sdk::clients::tangle::runtime::TangleConfig>);
+                            let mut instance = <#wrapper as gadget_sdk::event_listener::EventListener::<_, _>>::new(&ctx).await.expect("Failed to create event listener");
+                            let task = async move {
+                                if let Err(err) = gadget_sdk::event_listener::EventListener::<_, _>::execute(&mut instance).await {
+                                    gadget_sdk::error!("Error in event listener (now closing): {err}");
+                                }
+                            };
+                            gadget_sdk::tokio::task::spawn(task);
+                        }
+                    }
```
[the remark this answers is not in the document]
tbraun96: Fixed

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -87,30 +92,79 @@ pub(crate) fn job_impl(args: &JobArgs, input: &ItemFn) -> syn::Result<TokenStrea
...
+                    };
+
+                    let autogen_struct_name = if listener_str.contains("TangleEventListener") {
+                        let (_, _, struct_name) = generate_fn_name_and_struct(input);
+                        quote! {
+                            #struct_name
+                        }
+                    } else {
+                        unimplemented!("EvmEventListener not implemented yet");
+                    };
```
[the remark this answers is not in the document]
tbraun96: We can improve this iteratively. For now, it works, so let's leave it here until we find the better path forward.

tbraun96:
> There is still more work to be done here for sure, but I believe we are deviating from the original goal of this PR which is to support other even watchers other than the default ones (Substrate, EVM), and other types of Event Handlers, so why not adopting what we currently have and add more events watcher, like the common ones we see would be good for other blueprints?

My latest commit after your comment addresses this concern.

PR #338, opened 2024-10-07
fix: gadget proc macro tests

shekohex:
This PR is mainly to setup and enable the tests of `gadget-blueprint-proc-macro` crate in our CI, so that we do not break the proc-macro's codegen. We should extend the set of tests to cover each case we currently have.


drewstone:
@shekohex how do I execute these tests? I merged this into my `fix-eigenlayer-branch` but unsure how it was meant to be finished. Please advise.

shekohex:
> @shekohex how do I execute these tests? I merged this into my `fix-eigenlayer-branch` but unsure how it was meant to be finished. Please advise.

`cargo test -p gadget-blueprint-proc-macro`.

PR #339, opened 2024-10-08
feat: allow env vars for `ContextConfig` args

Serial-ATA:
Since the blueprint template now supports generating `Dockerfile`s, this makes it a little simpler.

For now, the blueprint template uses an entrypoint script to use env vars as arguments: https://github.com/webb-tools/blueprint-template/blob/main/docker/entrypoint.sh

This makes it possible to just have the binary as the entrypoint without having to rebuild the image when changing arguments.

### DIFF
--- a/sdk/src/config.rs
+++ b/sdk/src/config.rs
@@ -234,10 +234,10 @@ pub enum GadgetCLICoreSettings {
         test_mode: bool,
         #[structopt(long, short = "l", env)]
         log_id: Option<String>,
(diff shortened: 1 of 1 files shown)


drewstone:
Do the docs need updating? How will this be used? Can you add it to a subproject README or something?

Serial-ATA:
@drewstone Updated the comment for how it'll be used. This (most likely) isn't something that people will be interacting with unless they use the Dockerfile.

PR #340, opened 2024-10-08
Add documentation for listeners to gadget repo

tbraun96:


docs/event_listeners.md
```diff
@@ -0,0 +1,252 @@
...
+
+Next, implement `EventListener` for `Ticker`:
+
+```rust
+/// Implement the [`EventListener`] trait for the Ticker struct. [`EventListener`] has two type parameters:
+/// 
+/// - T: is the type of the event that the listener listens for. This can be anything that is Send + Sync + 'static.
+/// In this case, we are using [`Instant`], which is a timestamp.
+///
+/// - Ctx: is the context type that the listener receives when constructed. This can be anything that is Send + Sync + 'static,
```
drewstone, line 56: I feel this can be provided differently, even though I can mentally map `Ctx` to `MyContext` (because they're similar). I don't think it's connected super well though. This amount of inline comments should maybe be motivated outside of the code block.

drewstone (approved):

PR #346, opened 2024-10-14
Multi job runner + SDK main macro

tbraun96:


macros/blueprint-proc-macro/src/sdk_main.rs
```diff
@@ -0,0 +1,103 @@
...
+    let input = input.block.clone();
+
+    let tokens = quote! {
+        use gadget_sdk::tokio;
+        #[tokio::main #tokio_args]
+        async fn main() {
+            gadget_sdk::logging::setup_log();
+            // Load the environment and create the gadget runner
+            let config: gadget_sdk::config::ContextConfig = gadget_sdk::structopt::StructOpt::from_args();
+            let env = gadget_sdk::config::load(config.clone()).expect("Failed to load environment");
```
Serial-ATA, line 54: Why is there an optional `env` parameter to `sdk::main`? Is there a situation where you wouldn't want one? It gets loaded here anyway.

Being able to rename it would also be nice: `#[sdk::main(env = my_env)]`

drewstone:
CI also failing here still @tbraun96

blueprints/incredible-squaring/src/main.rs
```diff
@@ -8,139 +8,70 @@ use gadget_sdk::{
...
-        runner.register().await?;
-    }
-
-    info!("Running...");
-    // Run the gadget / AVS
-    runner.run().await?;
+    MultiJobRunner::new(&env)
+        .with_job_context(env.clone())
+        .with_registration(registration)
+        .add_job(x_square)
```
Serial-ATA, line 34: Naming nits:

`with_*` and `add_*` prefixes should be removed, builders typically just have `Builder::field()` methods. `with_*` is typically used for an associated function (like `MultiJobRunner::with_job_context(env).job(/* etc. */)`) ([example](https://rocket.rs/guide/v0.4/fairings/#attaching))

sdk/src/job_runner.rs
```diff
@@ -0,0 +1,145 @@
...
+    ///      signer: signer.clone(),
+    /// };
+    ///
+    /// let x_square2 = blueprint::XsquareEventHandler {
+    ///      service_id: self.env.service_id.unwrap(),
+    ///      client: client.clone(),
+    ///      signer,
+    /// };
+    ///
+    /// MultiJobRunner::new().with_job().add_job(x_square).with_job().add_job(x_square2).run().await;
```
Serial-ATA, line 102: Is that the current usage? If so, it's a little confusing. It'd probably be better to just pass in your own `JobBuilder` than have `with_job()` create one for you.

```rust
let runner = MultiJobRunner::new()
    .job(JobBuilder::new().job(x_square))
    .job(JobBuilder::new().registration(/* Something */).job(x_square2));
```

PR #355, opened 2024-10-16
fix: update sdk and blueprint utilities for tangle avs

Tjemmmic:
## Overview
As the Tangle AVS is being built, some additional changes and features are required in Gadget. 

### List of changes
- Fixes to Eigen keys in keystore
- Easier way to inject keys into a keystore for running tests
- Updates to executor to support output broadcast streams


sdk/src/utils/evm.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Register the account of the given ECDSA pair as an operator on EigenLayer.
+///
+/// # Returns
+/// - [`FixedBytes<32>`] - The transaction hash on success, otherwise an error
+///
+/// # Errors
+/// - [`Error::ElContract`] - If the registration call to the contract itself fails
+/// - [`Error::Keys`] - If the key pair fails to be converted for Alloy
+pub async fn register_as_operator(
```
drewstone, line 106: Where is this being used? Should this be the function used on the Eigenlayer runner? Is it just a dangling function for testing?

sdk/src/executor/process/utils.rs
```diff
@@ -32,25 +36,113 @@ macro_rules! craft_child_process {
...
+                            }
+                        }
+                        Err(e) => {
+                            crate::error!("Error reading from stdout: {}", e);
+                            break;
+                        }
+                    }
+                }
+                result = stderr_reader.next_line() => {
+                    crate::error!("READ TO STDERR: {:?}", result);
```
drewstone, line 79: These logs and the processing logic look repetitive to above, can we reuse it with a closure? i.e. the program flow from stdout to stderr looks identical, is it not?

sdk/src/executor/process/types.rs
```diff
@@ -73,123 +74,127 @@ impl GadgetProcess {
...
+                        if inbound_message.is_empty() {
+                            // Stream is completed - process is finished
+                            info!(
+                                "{} : STREAM COMPLETED - ENDING",
+                                self.process_name.to_string_lossy()
+                            );
+                            return ProcessOutput::Exhausted(messages);
+                        } else {
+                            // We received output from child process
+                            info!(
```
drewstone, line 183: Should these be info or debug logs? Can we do a cleanup of all the logging or at least let me know for sure everything is needed? It feels not like full mess, but it seems like some logs are in all caps which could lead me to believe they were for debugging?

blueprint-test-utils/src/helpers.rs
```diff
@@ -106,8 +104,10 @@ impl BlueprintProcessManager {
             std::path::absolute(&keystore_uri).expect("Failed to resolve keystore URI");
         let keystore_uri_str = format!("file:{}", keystore_uri_normalized.display());
 
-        let in_memory_keystore = FilesystemKeystore::open(keystore_uri_str.clone())
+        let filesystem_keystore = FilesystemKeystore::open(keystore_uri_str.clone())
```
drewstone, line 107: If we're using the filesystem keystore, are the keys being written to this when we generate? Are we doing proper cleanup after these tests execute or are developers locally building up a database of junk keys?

drewstone (changes requested):
Looks great, have some questions to address and then this is good to merge.

drewstone (approved):

blueprint-metadata/src/lib.rs
```diff
@@ -15,7 +15,7 @@ use rustdoc_types::{Crate, Id, Item, ItemEnum, Module};
 
 /// Generate `blueprint.json` to the current crate working directory next to `build.rs` file.
 pub fn generate_json() {
-    Config::builder().build().generate_json();
+    // Config::builder().build().generate_json();
```
drewstone, line 18: Is this mean to be here? @Tjemmmic

drewstone (changes requested):

sdk/src/config/context_config.rs
```diff
@@ -70,6 +70,9 @@ pub enum GadgetCLICoreSettings {
             required_if_eq("protocol", Protocol::Tangle.as_str())
         )]
         service_id: Option<u64>,
+        /// Whether to skip the registration process for a Blueprint
+        #[arg(long, env = "SKIP_REGISTRATION")]
+        skip_registration: Option<bool>,
```
shekohex, line 75: Why it is `Option<bool>` why not just `bool`? with default is `false`?

sdk/src/config/gadget_config.rs
```diff
@@ -198,15 +199,47 @@ impl<RwLock: lock_api::RawRwLock> GadgetConfiguration<RwLock> {
...
+    }
+
+    /// Returns the HTTP endpoint string from the bind address and bind port specified in the [`GadgetConfiguration`].
+    ///
+    /// # Note
+    /// This endpoint is *not* the same as the WS RPC endpoint in the [`GadgetConfiguration`]. This is the target endpoint
+    /// rather than the endpoint for the network that matches the specified [`Protocol`].
+    pub fn target_endpoint_http(&self) -> String {
+        format!("http://{}:{}", self.bind_addr, self.bind_port)
+    }
```
shekohex, line 234: This could be using `https` so I would really avoid this method and leave it to the end-users.

sdk/src/config/gadget_config.rs
```diff
@@ -198,15 +199,47 @@ impl<RwLock: lock_api::RawRwLock> GadgetConfiguration<RwLock> {
...
+
+    /// Returns the WS endpoint string from the bind address and bind port specified in the [`GadgetConfiguration`].
+    ///
+    /// # Note
+    /// This endpoint is *not* the same as the HTTP RPC endpoint in the [`GadgetConfiguration`]. This is the target endpoint
+    /// rather than the endpoint for the network that matches the specified [`Protocol`].
+    pub fn target_endpoint_ws(&self) -> String {
+        format!("ws://{}:{}", self.bind_addr, self.bind_port)
     }
```
shekohex, line 243: Same here.

sdk/src/config/protocol.rs
```diff
@@ -75,6 +75,8 @@ pub struct TangleInstanceSettings {
     pub blueprint_id: u64,
     /// The service ID for the Tangle blueprint instance
     pub service_id: u64,
+    /// Whether to skip the registration process
+    pub skip_registration: bool,
```
shekohex, line 79: Yeah you see, you already have this one here.

sdk/src/utils/evm.rs
```diff
@@ -0,0 +1,88 @@
...
+use alloy_network::EthereumWallet;
+use alloy_primitives::{Address, U256};
+use alloy_provider::{Provider, ProviderBuilder, RootProvider, WsConnect};
+use alloy_transport::BoxTransport;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    /// 1 day
+    pub static ref SIGNATURE_EXPIRY: U256 = U256::from(86400);
+}
```
shekohex, line 11: Remove `lazy_static` and use a constant? See: https://docs.rs/alloy/latest/alloy/primitives/struct.Uint.html#method.from_limbs

OR you can use https://doc.rust-lang.org/stable/std/cell/struct.LazyCell.html from the `std`.

shekohex (changes requested):
Looks good, left a few comments.

shekohex:
@Tjemmmic Do not forget to rebase too!

Issue #318, opened 2024-09-25
[TASK] Update tangle-subxt use of subxt/sp-core


tbraun96:
https://github.com/paritytech/subxt/issues/1783

After the above issue is addressed, we can upgrade our `subxt` in `tangle-subxt`, then from there, make a tangle PR, merge it, then open a gadget PR to use the latest tangle.

PR #361, opened 2024-10-18
fix(sdk)!: downgrade substrate dependencies for now

Serial-ATA:
It's currently impossible to use some APIs downstream, since they expose `sp_core` (v0.34.0) types instead of the `sdk::keystore::sp_core_subxt` (v0.31.0) types. Cargo will refuse to build blueprints using them. I just got rid of the `sdk::keystore::sp_core_subxt` hack and downgraded all of the dependencies necessary (I think).

This won't be an issue once #318 is taken care of.

I also re-exported `libp2p` for convenience.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1752,7 +1752,7 @@ dependencies = [
  "reqwest 0.12.8",
  "serde",
  "sha2 0.10.8",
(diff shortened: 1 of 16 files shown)


drewstone (approved):
Looks good

PR #368, opened 2024-10-22
fix(sdk): updated keystore support and fixes

Tjemmmic:
## Overview

Addresses/adds the following:
- Support for EigenLayer keys i.e., BLS BN254
- Fixes for BLS BN254, and BLS381
- Tests for all keystore types and key types


drewstone (commented):
Create a task for the CLI to create these keys, and add tests to it to verify the keys are created and accessible using the environment's keystore. Looks great!

drewstone (approved):

PR #382, opened 2024-10-23
feat(gadget-sdk): improve `MultiJobRunner` builder

Serial-ATA:
Jobs can be built using tuples now. Some comparisons:

### Old

```rust
MultiJobRunner::new(&env)
    .with_job()
    .with_default_price_targets()
    .finish(x_square)
```

### New

```rust
MultiJobRunner::new(&env)
    .job(x_square)
```

### Old

```rust
use tangle_subxt::tangle_testnet_runtime::api::runtime_types::tangle_primitives::services::PriceTargets;

let price_targets = PriceTargets {
    cpu: 5,
    mem: 10,
    storage_hdd: 15,
    storage_ssd: 20,
    storage_nvme: 25
};

MultiJobRunner::new(&env)
    .with_job()
    .with_price_targets(price_targets)
    .finish(x_square)
```

### New

```rust
use tangle_subxt::tangle_testnet_runtime::api::runtime_types::tangle_primitives::services::PriceTargets;

let price_targets = PriceTargets {
    cpu: 5,
    mem: 10,
    storage_hdd: 15,
    storage_ssd: 20,
    storage_nvme: 25
};

MultiJobRunner::new(env)
    .job((x_square2, price_targets))
```

closes #371

PR #383, opened 2024-10-23
fix(ci): add doc tests to CI

Serial-ATA:
`nextest` doesn't support doc tests currently. We need to make sure that examples in the docs compile.

**Needs #382 since it fixes another failing doc test**


Serial-ATA:
This'll keep failing while the `job` macro is being iterated on. I'll just make this a draft until it stops breaking.

drewstone:
@Serial-ATA I fixed up the docker docs tests here. The context doc test doesn't work yet.

PR #385, opened 2024-10-24
feat(cargo-tangle): key generation

Tjemmmic:
## Overview

Adds the following:

- Key generation to CLI
- Tests for generating keys via cargo-tangle commands


cli/README.md
```diff
@@ -103,3 +103,21 @@ The following environment variables are required for deploying the blueprint:
 export SIGNER="//Alice" # Substrate Signer account
 export EVM_SIGNER="0xcb6df9de1efca7a3998a8ead4e02159d5fa99c3e0d4fd6432667390bb4726854" # EVM signer account
 ```
+
+## Generating Keys from the Command Line
+
+The following command will generate a keypair for a given key type:
+
+```shell
+cargo tangle blueprint generate-keys -k <KEY_TYPE> -p <PATH> -s <SURI/SEED> --show-secret
```
drewstone, line 112: Would it benefit us to make this a command at the level of blueprint? What about generating keys here is subordinate to blueprint command?

drewstone (approved):

PR #394, opened 2024-10-24
fix(blueprints): remove blueprint metadata from eigen blueprint

Tjemmmic:
## Overview

Blueprint Metadata is not needed for the Eigenlayer Blueprints and is causing issues in the build script. This PR removes it from the incredible-squaring-eigenlayer blueprint.


shekohex (approved):
Looks good!

PR #398, opened 2024-10-24
feat(gadget-sdk): add utilities for interacting with Docker

### LINKED ISSUE #373: [TASK] Add Docker utils to the SDK
# Overview

The Obol blueprint at https://github.com/tangle-network/obol-dvt-blueprint heavily relies on spinning up a bunch of Docker containers with different commands. This is likely not the only service of its kind that we'd want to support.

The current utils in `src/docker.rs` are likely general enough to include in the SDK, will need to build more blueprints to see.

Serial-ATA:
This will be useful for multiple blueprints since interacting with Docker programmatically like this gets repetitive. It can also easily be extended in the future, when blueprints come to need extra features.

## Example

This is taken from the Obol blueprint:

```rust
let mut container = Container::new(&self.docker, IMAGE.to_string());

container
    .cmd(vec!["dkg", "--publish"])
    .binds(vec![format!("{}:{CHARON_DATA}", self.data_dir.display())]);

container.start(true).await?;
container.remove().await?;
```

**Also note that each new method has an example in the docs**

closes #373

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4778,6 +4778,7 @@ dependencies = [
  "auto_impl",
  "backon",
  "bincode",
+ "bollard",
  "ed25519-zebra 4.0.3",
  "eigensdk",
  "elliptic-curve",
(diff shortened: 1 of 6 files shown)


shekohex (approved):
LGTM

PR #429, opened 2024-11-01
feat: `blueprint-serde` crate

Serial-ATA:
This crates lets us convert to and from the `Field` type easily with `serde`'s `Serialize` and `Deserialize` traits. This will allow for custom types in job parameters, as well as making it easier overall to convert between these types.

The crate adds two public functions:

* `to_field` - Convert any `Serialize` type into a `Field`
* `from_field` - Attempt to convert a `Field` into a `DeserializeOwned` type

`Field::Struct` example:
```rust
use gadget_blueprint_serde::{new_bounded_string, BoundedVec, Field};
use serde::Serialize;

#[derive(Serialize)]
struct Person {
    name: String,
    age: u8,
}

let person = Person {
    name: String::from("John"),
    age: 40,
};

let expected = Field::Struct(
    new_bounded_string("Person"),
    Box::new(BoundedVec(vec![
        (
            new_bounded_string("name"),
            Field::String(new_bounded_string("John")),
        ),
        (new_bounded_string("age"), Field::Uint8(40)),
    ])),
);

let field = gadget_blueprint_serde::to_field(person).unwrap();
assert_eq!(expected, field);
```

While mostly useful for `Field::Struct`, one can also do:

```rust
let age: u8 = gadget_blueprint_serde::from_field(Field::Uint8(40)).unwrap();
```

Or conversion for any primitive type, if they wanted.


drewstone:
Will this be compatible with `no_std`?

Serial-ATA:
Yeah, that's the plan. I haven't made it `#![no_std]` yet just in case, but I don't imagine it'll need `std` at all.

drewstone:
We could have it be both too with the std feature. Just trying to keep us in the flow of doing so when we can.

Serial-ATA:
I can add an `std` feature just in case. I can't imagine us needing it though, none of the dependencies nor any of the de/serialization logic need `std`.

blueprint-serde/src/lib.rs
```diff
@@ -0,0 +1,104 @@
+#![cfg_attr(feature = "std", no_std)]
```
shekohex, line 1: Some top-level crate docs could be useful, with some examples and how to use them.

shekohex (approved):
Looks good overall, good job!

PR #453, opened 2024-11-06
feat: improved eigenlayer context and testing

Tjemmmic:
# Overview
This PR focuses on improvements to the EigenLayer blueprints through the following changes:
- Better, more reusable test utilities for EigenLayer blueprints
- Creates bindings for our common Eigenlayer Contracts to avoid issues with relative paths
- Moves the EigenLayer Incredible Squaring Blueprint test from the test-utils crate into the blueprint
- More methods automatically implemented via the Eigen Context
- Tests/examples for the context functions


blueprints/examples/contracts/src/ContextExampleTaskManager.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.13;
+
+contract ContextExampleTaskManager {
```
drewstone, line 4: This name is weird, just `ExampleTaskManager` is better.

blueprints/examples/src/examples/eigen_context.rs
```diff
@@ -0,0 +1,196 @@
...
+        })
+        .unwrap_or_else(|_| address!("0000000000000000000000000000000000000000"));
+
+    let context_example_task_manager =
+        ContextExampleTaskManager::ContextExampleTaskManagerInstance::new(
+            context_example_address,
+            get_provider_http(&env.http_rpc_endpoint),
+        );
+
+    Ok(DemonstrateEigenlayerContextEventHandler {
```
drewstone, line 46: Should use the `new(...)` method now and migrate what you can over to that.

blueprints/examples/src/examples/eigen_context.rs
```diff
@@ -0,0 +1,196 @@
...
+    id = 0,
+    params(event, log),
+    event_listener(
+        listener = EvmContractEventListener<ContextExampleTaskManager::NewTaskCreated>,
+        instance = ContextExampleTaskManager,
+        abi = CONTEXT_EXAMPLE_TASK_MANAGER_ABI_STRING,
+        pre_processor = handle_events,
+    ),
+)]
+pub async fn demonstrate_eigenlayer_context(
```
drewstone, line 65: Function names don't need to be super absolutely tied to the example, it's more readable to share in docs that this is a "handle_task" function, not a `demonstrate_eigenlayer_context`, which holds little meaning in our job terminology.

blueprints/examples/src/main.rs
```diff
@@ -1,16 +1,42 @@
...
-        .await?;
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    info!("~~~ Executing Blueprint Examples ~~~");
+
+    // Read the EXAMPLE_PROTOCOL environment variable
+    let example_protocol = env::var("EXAMPLE_PROTOCOL")
+        .unwrap_or_else(|_| "tangle".to_string())
+        .to_lowercase();
+
+    match example_protocol.as_str() {
```
drewstone, line 17: Awesome!

blueprints/examples/src/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+use gadget_sdk::runners::eigenlayer::EigenlayerConfig;
+use gadget_sdk::runners::BlueprintRunner;
+use gadget_sdk::utils::evm::get_provider_http;
+use reqwest::Url;
+use std::net::IpAddr;
+use std::path::Path;
+use std::str::FromStr;
+
+#[tokio::test]
+async fn test_demonstrate_eigenlayer_context() {
```
drewstone, line 23: The naming is very poor. The test eigenlayer context can test a `handle_task` function. It's ok! Make it clear and human understandable. We're putting these things in the docs and communicating with humans.
```suggestion
async fn test_eigenlayer_context() {
```

blueprints/examples/src/tests.rs
```diff
@@ -0,0 +1,103 @@
...
+    inject_test_keys(&keystore_uri, KeyGenType::Anvil(1))
+        .await
+        .expect("Failed to inject testing keys for Blueprint Examples Test");
+    let keystore_uri_normalized =
+        std::path::absolute(&keystore_uri).expect("Failed to resolve keystore URI");
+    let keystore_uri_str = format!("file:{}", keystore_uri_normalized.display());
+
+    // Create the GadgetConfiguration
+    let config = ContextConfig {
+        gadget_core_settings: GadgetCLICoreSettings::Run {
```
drewstone, line 64: I suggest creating methods to craft the Eigenlayer specific settings (and other protocols) so that this isn't the finished product. There are blueprint_id/service_id in this that shouldn't be expected to be supplied by end developers. Same with Symbiotic contract data.

drewstone: i.e. `let config = ContextConfig::create_eigenlayer_config(...)`

Also why is blueprint/service id even `Some`? Shouldn't it be `None`?

blueprints/incredible-squaring-eigenlayer/src/tests.rs
```diff
@@ -0,0 +1,344 @@
...
+        );
+    }
+}
+
+/// Sets up the test environment for the EigenLayer Blueprint.
+///
+/// # Description
+/// - Sets all the necessary environment variables for the necessary EigenLayer Contract Addresses.
+/// - Returns a [`EigenlayerTestEnvironment`] struct containing the test environment state.
+pub async fn setup_eigenlayer_test_environment(
```
drewstone, line 136: Why was this function specifically moved out of test utils? Is it not something to be reused across all eigen blueprint tests?

drewstone (changes requested):
Looks good, some changes to fix and improve. Naming needs improvement, it feels robotic. Some test util function for eigen environment should not have been removed IMO as it's a reusable eigen environment system.

blueprint-metadata/src/lib.rs
```diff
@@ -15,7 +15,7 @@ use rustdoc_types::{Crate, Id, Item, ItemEnum, Module};
 
 /// Generate `blueprint.json` to the current crate working directory next to `build.rs` file.
 pub fn generate_json() {
-    Config::builder().build().generate_json();
+    // Config::builder().build().generate_json();
```
drewstone, line 18: Should this still be commented out @Tjemmmic ?

blueprints/examples/src/tests.rs
```diff
@@ -0,0 +1,136 @@
+use crate::eigen_context;
+use crate::eigen_context::ExampleTaskManager;
+use alloy_provider::Provider;
+use blueprint_test_utils::helpers::get_receipt;
+use blueprint_test_utils::incredible_squaring_helpers::start_default_anvil_testnet;
```
drewstone, line 5: Why is this apart of the incredible squaring helpers and not some `evm_test_utils`? Seems like the organization can be put in a properly named file. Incredible squaring stuff should be explicit to incredible squaring. Otherwise it's more general and reusable.

blueprints/examples/src/tests.rs
```diff
@@ -0,0 +1,136 @@
...
+        "EXAMPLE_TASK_MANAGER_ADDRESS",
+        context_example_task_manager_address.to_string(),
+    );
+    let context_example_task_manager =
+        ExampleTaskManager::new(context_example_task_manager_address, provider.clone());
+
+    info!("Starting Eigenlayer Blueprint Context Test...");
+
+    // Set up Task Spawner
+    let _task_spawner_handle = tokio::task::spawn(async move {
```
drewstone, line 51: If this is not being used can you remove it? I had assumed we are using the task spawner from the `incredible_squaring_helpers`

blueprints/incredible-squaring-eigenlayer/src/tests.rs
```diff
@@ -0,0 +1,267 @@
...
+    tokio::spawn(async move {
+        response_listener.await;
+    });
+
+    info!("Starting Blueprint Binary...");
+
+    let blueprint_process_manager = BlueprintProcessManager::new();
+    let current_dir = std::env::current_dir().unwrap();
+    let xsquare_task_program_path = PathBuf::from(format!(
+        "{}/../../target/release/incredible-squaring-blueprint-eigenlayer",
```
drewstone, line 83: I think the test inside this blueprint should not execute the binary it should just test running it directly. Otherwise this is forcing the user to build release mode BEFORE running the test. If we think this should be here I'd suggest maybe using debug mode? Is there some pre-build process we can call to enforce this binary actually exists with proper error handling if we keep this?

drewstone: Chat with your teammates about what we should do here, you don't need to agree with me.

drewstone (approved):
I'm not even sure `incredible_squaring_helpers` is tied to the incredible squaring anymore? Seems it just exports anvil? Can we rename it? Little bit of cleanup or organization missing before this can be totally merged. I will approve nonetheless so you can fix up, get CI passed, and merge if I'm asleep.

blueprints/examples/src/tests.rs
Tjemmmic: The task spawner for the incredible squaring blueprint is different and utilizes the Incredible Squaring Task Manager. This is just a very simple version for the example blueprint.

blueprints/incredible-squaring-eigenlayer/src/tests.rs
Tjemmmic: I think everyone (including me) agrees that we shouldn't test with the binary. I was going to make this change in a subsequent PR since the  changes keep adding up here. But I went ahead and made those changes to get it resolved.

PR #469, opened 2024-11-12
feat(gadget-sdk)!: integrate `blueprint-serde`

Serial-ATA:
This gets rid of the manual conversions does with the `ValueIntoFieldType` trait and `field_type_to_param_token` utils, letting all conversions be unified under `blueprint-serde`.

This doesn't support custom struct args yet, that's a future (more complicated) issue.


drewstone:
Merge conflicts here @Serial-ATA and then g2g

Issue #407, opened 2024-10-25
[BUG] Failed to decode the field


shekohex:
I get the following error when trying the hello world blueprint.

```
 INFO gadget: Pre-processing event for sid/bid = 0/0 ...
 INFO gadget: Pre-processing event for sid/bid = 0/0 ...
 INFO gadget: Pre-processing event for sid/bid = 0/0 ...
 INFO gadget: Found actionable event for sid/bid = 0/0 ...
thread 'tokio-runtime-worker' panicked at src/lib.rs:17:1:
Failed to decode the field
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
ERROR gadget: An Event Handler for SayHelloEventHandler has stopped running
ERROR gadget: Job 0 exited prematurely
Error: Other("Error in Event Handler for SayHelloEventHandler: RecvError(())")
```

## Steps

1. Create the blueprint using the blueprint template
2. Build and deploy the blueprint using tangle cli
3. Register on the blueprint using any account (ex //Bob)
4. Request an instance using another account (ex //Charlie)
5. Approve the request from Bob
6. Dispatch a job call for job#0 without any parameters using Charlie
7. you will get the above error.

## Expected

When I get invalid job args, or empty, the worker should not stop and instead recover gracefully without panicking!

shekohex:
![image](https://github.com/user-attachments/assets/79a5029c-1087-4b6f-b050-6d706bae8369)

Issue #381, opened 2024-10-23
[TASK] proc-macros should not panic


shekohex:
Our proc-macros should not panic, and instead use proper error handling. 

* Missing a parameter? Return syn::Error with the context of the input attached. This would provide way better diagnostics in the IDE. 

* Provide hints and help, guys we are building tools for the developers, lets make the DX as best as we can.

PR #472, opened 2024-11-12
refactor(macros): cleanup macros, add better error handling, dedup code, DX

tbraun96:
Addresses #407 and #381


macros/blueprint-proc-macro/tests/tests.rs
```diff
@@ -1,22 +1,18 @@
...
-    //     t.pass("tests/valid_cases/job/*.rs");
-    // }
-
     #[test]
     fn test_jobs_invalid_cases() {
         let t = TestCases::new();
         t.compile_fail("tests/invalid_cases/job/*.rs");
     }
-
-    #[test]
```
drewstone, line 17: The valid test cases you fixed below aren't running now ya? Can we make them run? Or it's simply enough to define them above and ensure they compile?

macros/playground/src/lib.rs
```diff
@@ -231,74 +231,43 @@ mod tests {
...
         event_listener(
-            listener = PeriodicEventListener<1500, WebPoller, serde_json::Value, Arc<AtomicUsize>>,
+            listener = PeriodicEventListener<1500, WebPollerTest, serde_json::Value, Arc<AtomicUsize>>,
             pre_processor = pre_process,
             post_processor = post_process,
         ),
     )]
     // Maps a boolean value obtained from pre-processing to a u8 value
-    pub async fn web_poller(value: bool, count: Arc<AtomicUsize>) -> Result<u8, Infallible> {
+    pub async fn web_poller_test(value: bool, count: Arc<AtomicUsize>) -> Result<u8, Infallible> {
```
drewstone, line 241: Can't we test this in the blueprint examples repo? It's getting to be a bit of a nested mess IMO. PeriodicListener with WebPoller and then another PeriodicListener with a WebPollerTest containing a WebPoller.

I would suggest getting CI to test the example blueprint instead? Would that accomplish your goal here? It would also be good to have a PeriodicListener example as a simple cron-job no inner listener too, no complexity, just "here's how you trigger your job every 2 seconds".

drewstone (commented):
Looks good, some comments and would appreciate more reviews on this @Serial-ATA @shekohex

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -278,7 +255,7 @@ pub(crate) fn generate_event_workflow_tokenstream(
                     // If is_raw, assume the actual context is the second param
                     quote! { ctx. #field_in_self .clone() }
                 })
-                .expect("No context found");
+                .ok_or_else(|| syn::Error::new(Span::call_site(), "Must specify a context"))?;
```
shekohex, line 259: you should use `params` here for the instead of the `Span::call_site()` so the errors appears on the function args.

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -479,7 +461,7 @@ pub(crate) fn get_event_handler_args<'a>(
     let event_handler_args = x.difference(&y).copied().collect::<Vec<_>>();
     let event_handler_types = event_handler_args
         .iter()
-        .map(|r| param_types.get(*r).unwrap())
+        .map(|r| param_types.get(*r).expect("Should exist"))
         .collect::<Vec<_>>();
```
shekohex, line 465: Map to proper errors instead of expect/panic here then collect into a `Result<Vec<_>>()` for proper error handling in this case.

macros/blueprint-proc-macro/src/report.rs
```diff
@@ -27,6 +27,7 @@ mod kw {
     syn::custom_keyword!(evm);
     syn::custom_keyword!(event_listener);
     syn::custom_keyword!(skip_codegen);
+    syn::custom_keyword!(description);
```
shekohex, line 30: This is not needed, see next for why it is not need.

macros/blueprint-proc-macro/src/report.rs
```diff
@@ -74,7 +75,7 @@ pub(crate) fn report_impl(args: &ReportArgs, input: &ItemFn) -> syn::Result<Toke
     let report_def = ReportDefinition {
         metadata: ReportMetadata {
             name: fn_name_string.clone().into(),
-            description: None, // TODO: Add support for description
+            description: args.description.as_ref().map(|r| r.as_str().into()),
```
shekohex, line 78: This one actually was not a TODO. This field will get updated by `blueprint-metdata` crate during build time. Hence, if we added this one here, it will be overwritten by the build step; I recommend leaving it with `None` and just add a comment explaining why.

macros/blueprint-proc-macro/src/report.rs
```diff
@@ -209,6 +209,7 @@ pub(crate) struct ReportArgs {
     /// Optional: Verifier for the report result, currently only supports EVM verifier.
     /// `#[report(verifier(evm = "MyVerifierContract"))]`
     verifier: Verifier,
+    description: Option<String>,
```
shekohex, line 212:
```suggestion
```

macros/blueprint-proc-macro/src/report.rs
```diff
@@ -284,12 +286,18 @@ impl Parse for ReportArgs {
...
             } else if lookahead.peek(kw::skip_codegen) {
                 let _ = input.parse::<kw::skip_codegen>()?;
                 skip_codegen = true;
             } else if lookahead.peek(Token![,]) {
                 let _ = input.parse::<Token![,]>()?;
+            } else if lookahead.peek(kw::description) {
+                let _ = input.parse::<kw::description>()?;
+                let _ = input.parse::<Token![=]>()?;
+                let type_str: LitStr = input.parse()?;
+                description = Some(type_str.value())
```
shekohex, line 300:
```suggestion
```

macros/blueprint-proc-macro/src/report.rs
```diff
@@ -228,8 +229,9 @@ impl Parse for ReportArgs {
         let mut job_id = None;
         let mut interval = None;
         let mut metric_thresholds = None;
+        let mut description = None;
```
shekohex, line 232:
```suggestion
```

shekohex (changes requested):
Overall Looks good! 

aside from the removed test cases, we need to expand more and add a test case for each scenario (AI Could be useful here for generating different scenarios); Testing here is very important to make sure changes to the macros or other things does not break the SDK.

macros/blueprint-proc-macro/tests/tests.rs
tbraun96: Before, they weren't technically running, only checking for syntactical correctness. Now, it does the same thing but implicitly with being attached to the module tree. We could make them run once to check input/output correctness, though, maybe in another PR since it would be beyond the scope of this task.

macros/playground/src/lib.rs
tbraun96: We currently don't have a clean way of testing blueprints unless we inject test-related code into them (like I did for the `web_poller_test`). While we could test the one inside `examples`, we would have to add more code to it, confusing the user by mixing prod code with test code. At least with `web_poller_test`, I just need to import the already-existing code inside `examples`, then add a thin wrapper over it where I am then able to inject the test-related code.

macros/blueprint-proc-macro-core/src/lib.rs
```diff
@@ -45,6 +45,8 @@ pub enum FieldType {
     List(Box<FieldType>),
     /// A Struct of items of type [`FieldType`].
     Struct(String, Vec<(String, Box<FieldType>)>),
+    /// Tuple
+    Tuple(Vec<FieldType>),
```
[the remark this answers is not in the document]
tbraun96: https://github.com/tangle-network/tangle/issues/819

macros/blueprint-proc-macro/src/job.rs
```diff
@@ -255,7 +255,9 @@ pub(crate) fn generate_event_workflow_tokenstream(
                     // If is_raw, assume the actual context is the second param
                     quote! { ctx. #field_in_self .clone() }
                 })
-                .ok_or_else(|| syn::Error::new(Span::call_site(), "Must specify a context"))?;
+                .ok_or_else(|| {
+                    syn::Error::new_spanned(quote! { (#(#params),*) }, "Must specify a context")
+                })?;
```
shekohex, line 260: This not what I meant when I mentioned using `params` instead of `call_site`. params may be already carrying the original span where it has been defined, if not (i.e we created these params somewhere without spans) we should pass to this function `get_event_handler_args` a span, from the original declaration of the job macro.

If something like this not possible, or hard to do, using call_site here is _okay_

tbraun96: Yeah, `params` does not work. I have to wrap it as seen above.

shekohex: This won't give you the correct location of the error, it will mark the whole job as an error.

shekohex (approved):

PR #474, opened 2024-11-13
feat(gadget-blueprint-serde)!: change array/tuple serialization

Serial-ATA:
Previously, `Vec<T>`, `[T; N]`, and `(T, T, T)` would all serialize to `Field::List`. Now the following happens:

* `Vec<T>` -> `Field::List`
* `[T; N]` **OR** `(T, T, T)` -> `Field::Array`
  * The major change is that homogeneous tuples are treated as arrays now
* `(X, Y, Z)` -> Error (for now), waiting for tuple support in the runtime.


shekohex (approved):
LGTM!

PR #476, opened 2024-11-15
feat: Add utils for blueprints: hashing, config convenience functions

tbraun96:


sdk/src/config/gadget_config.rs
```diff
@@ -129,6 +130,32 @@ impl<RwLock: lock_api::RawRwLock> Default for GadgetConfiguration<RwLock> {
 }
 
 impl<RwLock: lock_api::RawRwLock> GadgetConfiguration<RwLock> {
+    /// Returns a libp2p-friendly identity keypair.
+    pub fn libp2p_identity(&self) -> Result<libp2p::identity::Keypair, Error> {
+        let ed25519 = *self.first_ed25519_signer()?.signer();
```
drewstone, line 135: If this doesn't exist should we generate one? Should we keep a separate key explicitly for libp2p connections in the event the environment is actually using this key for signing txes or other purposes?

drewstone (approved):
Looks good! Comment on dealing with keys for libp2p

sdk/src/config/gadget_config.rs
drewstone: Will merge this but please follow up! @tbraun96

PR #477, opened 2024-11-16
feat: add more service ctx methods for Tangle

drewstone:


shekohex (commented):
Nit; can we drop the `get_` prefix from method names? it does not add any meaningful to the operation, which is clearly retrieving something, I do try hard as much as I can to avoid these getter and setters to the method name, and let the that be already inferred from the method signature.
 
Otherwise, Looks good overall.

PR #497, opened 2024-11-22
feat: add MPCContext derive + test utils refactor

### LINKED ISSUE #495: [TASK] MPC refactor: add common functions to context derive
Now that we've built multiple MPCs, duplicated code can be abstracted away inside the SDK and test utils.

Various MPC-related functions could be added as auto-generated context functions

e.g.,

```rust
#[derive(Mpc)]
struct MyContext { // ... }
// ...
fn my_job(context: MyContext) {
    let (i, operators) = context.get_operators_and_self_party_index();
    // ...
}
```

As for testing, we can have the initialization handle much of the repetitive boilerplate. Afterward, there may be room for a macro to easily define tests, especially MPC-specific macros that often store keys locally using a call ID, and that call ID needs to be funneled into the next job (e.g., keygen -> signing) in order to retrieve data from the local store.

tbraun96:
Closes #495

### DIFF
--- a/blueprint-test-utils/src/tangle/mod.rs
+++ b/blueprint-test-utils/src/tangle/mod.rs
@@ -23,7 +23,8 @@ pub fn run() -> Result<SubstrateNode, Error> {
 }
 
 #[macro_export]
-macro_rules! generic_test_blueprint {
+/// A template that makes creating domain-specific macros for tangle-based blueprints easier
+macro_rules! tangle_blueprint_test_template {
     (
         $blueprint_path:expr,
         $N:tt,
@@ -92,7 +93,7 @@ macro_rules! test_tangle_blueprint {
         [$($inputs:expr),+],
         [$($expected_output:expr),+]
     ) => {
-        generic_test_blueprint!(
+        tangle_blueprint_test_template!(
             $blueprint_path,
             $N,
             |client, handles, blueprint| async move {
--- a/blueprint-test-utils/src/tangle/transactions.rs
+++ b/blueprint-test-utils/src/tangle/transactions.rs
@@ -27,7 +27,7 @@ pub async fn create_blueprint(
     Ok(())
 }
 
-pub async fn join_delegators(
+pub async fn join_operators(
     client: &TestClient,
     account_id: &TanglePairSigner<sp_core::sr25519::Pair>,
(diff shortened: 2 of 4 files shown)


sdk/src/ctx.rs
```diff
@@ -305,3 +306,64 @@ pub trait EigenlayerContext {
...
+
+use crate::network::NetworkMultiplexer;
+use color_eyre::Result;
+use round_based::PartyIndex;
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+/// `MPCContext` trait provides access to MPC (Multi-Party Computation) functionality from the context.
+#[async_trait::async_trait]
+pub trait MPCContext {
```
drewstone, line 318: Suggest we turn `contexts` into a folder in the SDK and separate each one into its own file now. There's plenty of contexts that this would be worth doing in this PR.

drewstone: I would also suggest creating a P2P network context that provides the general p2p network setup too. It doesn't need to do anything with MPC and it would be beneficial for me to add to the docs.

drewstone (changes requested):

blueprint-test-utils/src/tangle/transactions.rs
```diff
@@ -0,0 +1,233 @@
...
+    let call = api::tx().services().create_blueprint(blueprint);
+    let res = client
+        .tx()
+        .sign_and_submit_then_watch_default(&call, account_id)
+        .await?;
+    wait_for_in_block_success(res).await?;
+    Ok(())
+}
+
+pub async fn join_delegators(
```
drewstone, line 30: Join operators?

blueprint-test-utils/src/tangle/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+        .arg_val("rpc-cors", "all")
+        .arg_val("rpc-methods", "unsafe")
+        .arg("rpc-external")
+        .arg_val("sealing", "manual")
+        .clone();
+    builder.spawn()
+}
+
+#[macro_export]
+macro_rules! generic_test_blueprint {
```
drewstone, line 26: What's the difference between this macro and test tangle blueprint macro? The naming is not very clear what generic means here.

PR #507, opened 2024-11-29
feat: allow zero-sized inputs for tangle-based macros

tbraun96:


drewstone (approved):

drewstone:
Is this something that all our MpC blueprints can use? Is there a world it's not just a hack / extension that should live outside the SDK as a library extension?

PR #533, opened 2024-12-07
Call ID Insertion and Resolution For #520

### LINKED ISSUE #520: [TASK] Testing fetcher is not automatically generated
In the presence of another source (ex. GitHub releases), `blueprint-metadata` will not generate a testing fetcher for a blueprint:

https://github.com/tangle-network/gadget/blob/726181454e753d36b27d92714453bfcf18f76848/blueprint-metadata/src/lib.rs#L319-L333

When a testing fetcher is not present, and one attempts to test, they will get **stuck in an infinite loop**, as it is not a hard error:

https://github.com/tangle-network/gadget/blob/726181454e753d36b27d92714453bfcf18f76848/blueprint-manager/src/executor/event_handler.rs#L267-L274

### LINKED ISSUE #534: Runtime Call ID Injection (for the job function)
Tangle Listeners should require context structs to contain the `call_id` field. That way, at runtime, the call_id can be dynamically injected. By doing this, we prevent race conditions.

```
#[job(
    id = 0,
    params(x),
    event_listener(
        listener = TangleEventListener<MyContext, JobCalled>,
        pre_processor = services_pre_processor,
        post_processor = services_post_processor,
    ),
)]
/// Returns x^2 saturating to [`u64::MAX`] if overflow occurs.
pub fn xsquare(x: u64, context: MyContext) -> Result<u64, gadget_sdk::Error> {
    assert!(context.call_id.is_some());
    Ok(x.saturating_pow(2))
}
```

This task does not address the injection of the call_id into the post-processor. To achieve that, a different approach will be required. For now, providing the call ID to the job function has the immediate advantage of giving related tasks between nodes the ability to create symmetric hashes.

tbraun96:
Resolves #520
Resolves #534 

1. Define the context. `TangleClientContext` now requires a call_id field:

```rust
#[derive(Clone, TangleClientContext)]
pub struct MyContext {
    #[config]
    sdk_config: StdGadgetConfiguration,
    #[call_id]
    call_id: Option<u64>,
}
```

2. Access the call id at runtime via the context:

```rust
#[job(
    id = 0,
    params(x),
    event_listener(
        listener = TangleEventListener<MyContext, JobCalled>,
        pre_processor = services_pre_processor,
        post_processor = services_post_processor,
    ),
)]
/// Returns x^2 saturating to [`u64::MAX`] if overflow occurs.
pub fn xsquare(x: u64, context: MyContext) -> Result<u64, gadget_sdk::Error> {
    assert!(context.call_id.is_some()); // <----- Assertion success; runtime injection occurs
    Ok(x.saturating_pow(2))
}

```

NOTE: This only propagates the call_id from the pre-processor to the job function. The propagation from the job function to the post-processor is not covered.


blueprints/incredible-squaring/contracts/src/IncredibleSquaringBlueprint.sol
```diff
@@ -57,7 +56,7 @@ contract IncredibleSquaringBlueprint is BlueprintServiceManagerBase {
         bytes calldata participant,
         bytes calldata inputs,
         bytes calldata outputs
-    ) public payable override {
```
drewstone, line 60: Better to actually look at what needs to be overriden and make that change.

macros/blueprint-proc-macro/src/special_impls/tangle.rs
```diff
@@ -91,37 +92,50 @@ pub(crate) fn get_tangle_job_processor_wrapper(
         const PARAMETER_COUNT: usize = #parameter_count;
     };
 
+    // let non_job_param_map = get_non_job_arguments(param_map, job_params);
```
drewstone, line 95: Needed?

drewstone (changes requested):

macros/playground/src/lib.rs
```diff
@@ -26,16 +27,21 @@ impl std::fmt::Display for Error {
...
 
 impl std::error::Error for Error {}
-#[derive(Copy, Clone)]
-pub struct MyContext;
+#[derive(Clone, TangleClientContext)]
+pub struct MyContext {
+    #[config]
+    pub config: StdGadgetConfiguration,
+    #[call_id]
+    pub call_id: Option<u64>,
```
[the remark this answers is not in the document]
Serial-ATA: The name must be `call_id`

Serial-ATA (approved):

PR #547, opened 2024-12-13
Schnorr MuSig Debugging + SDK improvements + SDK debugging

tbraun96:


sdk/src/contexts/mpc.rs
```diff
@@ -1,12 +1,11 @@
...
 use std::collections::BTreeMap;
-use std::sync::Arc;
+use subxt_core::tx::signer::Signer;
 use subxt_core::utils::AccountId32;
+use crate::contexts::{ServicesContext, TangleClientContext};
 
 /// `MPCContext` trait provides access to MPC (Multi-Party Computation) functionality from the context.
 #[async_trait::async_trait]
-pub trait MPCContext {
+pub trait MPCContext: ServicesContext<Config = <Self as TangleClientContext>::Config> + TangleClientContext {
```
drewstone, line 8: Can you create a MPCClient and just have this context return this client? This way we can debug the impl a lot more easily?

sdk/src/contexts/mpc.rs
```diff
@@ -16,48 +15,166 @@ pub trait MPCContext {
...
+            .tangle()
+            .map(|c| c.blueprint_id)
+            .map_err(|err| crate::color_eyre::Report::msg(format!("Blueprint ID not found in configuration: {err}")))
+    }
+
+    /// Retrieves the current call ID for this job
+    ///
+    /// # Errors
+    /// Returns an error if failed to retrieve the call ID from storage
+    #[deprecated(note = "Use `context.call_id` field instead. This function is highly prone to race conditions")]
```
drewstone, line 62: Why are we putting deprecated code in when we don't really have any blueprints live? Just clean things up, accelerate towards the goal.

sdk/src/contexts/mpc.rs
```diff
@@ -16,48 +15,166 @@ pub trait MPCContext {
...
+        crate::Error,
+    > {
+        Ok(self.get_party_index_and_operators().await?.1.into_iter().enumerate().map(|(i, (id, _))| (i as _, id)).collect())
+    }
+
+    /// Retrieves the current blueprint ID from the configuration
+    ///
+    /// # Errors
+    /// Returns an error if the blueprint ID is not found in the configuration
+    fn blueprint_id(&self) -> crate::color_eyre::Result<u64> {
```
drewstone, line 50: This seems like it should be on the services context/client, CC @Tjemmmic

drewstone (changes requested):
@Tjemmmic @tbraun96 please meet, since we're changing these contexts let's accelerate towards creating clients for them instead and making the contexts return these clients. It'll be a lot more easily to implement / debug rather than implementing inside a macro.

sdk/src/contexts/mpc.rs
Tjemmmic: Yeah I moved a bunch of functions in the new contexts crate

sdk/src/contexts/gossip_network.rs
```diff
@@ -1,5 +1,20 @@
...
 pub trait GossipNetworkContext {
-    /// Get the Goossip client from the context.
-    fn gossip_network(&self) -> &crate::network::gossip::GossipHandle;
+    /// Creates a network delivery wrapper for MPC communication
+    fn create_network_delivery_wrapper<M>(
+        &self,
+        mux: std::sync::Arc<crate::network::NetworkMultiplexer>,
+        party_index: crate::round_based::PartyIndex,
+        task_hash: [u8; 32],
+        parties: std::collections::BTreeMap<crate::round_based::PartyIndex, crate::subxt_core::ext::sp_core::ecdsa::Public>,
```
drewstone, line 11: Need to generalize over sp_core::ecdsa::Public

drewstone:
What's the status of this?

sdk/src/contexts/services.rs
```diff
@@ -1,136 +1,438 @@
...
+        operators: Vec<crate::ext::subxt::utils::AccountId32>,
+    ) -> Result<
+        Vec<(
+            crate::ext::subxt::utils::AccountId32,
+            crate::tangle_subxt::tangle_testnet_runtime::api::runtime_types::pallet_multi_asset_delegation::types::operator::OperatorMetadata<
+                crate::ext::subxt::utils::AccountId32,
+                crate::ext::tangle_subxt::tangle_testnet_runtime::api::assets::events::burned::Balance,
+                crate::ext::tangle_subxt::tangle_testnet_runtime::api::assets::events::accounts_destroyed::AssetId,
+                crate::ext::tangle_subxt::tangle_testnet_runtime::api::runtime_types::tangle_testnet_runtime::MaxDelegations,
+                crate::ext::tangle_subxt::tangle_testnet_runtime::api::runtime_types::tangle_testnet_runtime::MaxOperatorBlueprints,
```
Serial-ATA, line 100:
```suggestion
            AccountId32,
            OperatorMetadata<
                AccountId32,
                Balance,
                AssetId,
                MaxDelegations,
                MaxOperatorBlueprints,
```
All these long type paths can be removed since it's no longer in a macro (this is true for all methods)

Serial-ATA:
Outdated due to workspace migration

PR #625, opened 2025-01-30
feat: blueprint pre-registration

Tjemmmic:
# Overview

Adds the ability to pre-register for blueprints i.e., running a blueprint registers and then exits with the registration success status the first time it is run without already having been registered.

Also adds pre-registration compatibility to the test harnesses and pre-registration tests to blueprints to ensure this flow works as intended.


Tjemmmic:
Will need to be updated once https://github.com/tangle-network/gadget/pull/626 is merged.

PR #626, opened 2025-01-30
feat: add dynamic multinode test executor

### LINKED ISSUE #619: [TASK] Multinode Test Harness w/ Dynamic add/drop support
Our test harness currently works for `N=1` nodes. For testing blueprints that only need 1 node, what we currently have works well and follows a composable design (as opposed to a "black box" design we use to have).

Now, we need to get `N=n` nodes working. The same composability should still apply, and should be reverse-compatible with existing tests to not cause interruption.

Additionally, we should add the ability to add and drop nodes on command.

tbraun96:
Closes #619


crates/testing-utils/tangle/src/harness.rs
```diff
@@ -12,43 +12,405 @@ use crate::{
...
+                            >,
+                        > + Send
+                        + 'static,
+                >,
+            > + Send
+            + Sync
+            + 'static,
+    > JobCreator for T
+{
+}
```
Tjemmmic, line 116: We could use a few reusable types here to reduce complexity and increase readability

Tjemmmic: Perhaps something like

```
type EventHandlerBox = Box<dyn InitializableEventHandler + Send + 'static>;
type JobResult = Result<EventHandlerBox, RunnerError>;
type JobFuture = Pin<Box<dyn Future<Output = JobResult> + Send + 'static>>;

trait JobCreator: Fn(GadgetConfiguration) -> JobFuture + Send + Sync + 'static {}

impl<T> JobCreator for T 
where
    T: Fn(GadgetConfiguration) -> JobFuture + Send + Sync + 'static
{}
```

crates/testing-utils/tangle/src/harness.rs
```diff
@@ -12,43 +12,405 @@ use crate::{
...
+    /// completed or the `MultiNodeTangleTestEnv` is dropped.
+    ///
+    /// The `MultiNodeTangleTestEnv` provides methods for starting and stopping jobs. The `start_job`
+    /// method takes a job ID and a closure that returns an `InitializableEventHandler`. The closure
+    /// will be called with the `GadgetConfiguration` for the specified job ID, and the returned event
+    /// handler will be used to execute the job. The `stop_job` method takes a job ID and will signal
+    /// the background task to stop the job with the specified ID.
+    ///
+    /// The `MultiNodeTangleTestEnv` also provides a method for getting the `GadgetConfiguration` for
+    /// a given job ID. This can be used to get the configuration for a job without starting the job.
```
Tjemmmic, line 137: This is a lot, it might be helpful to have some of the items link to themselves for clarity i.e., `[`start_job`](MultiNodeTangleTestEnv::start_job)`

crates/testing-utils/tangle/src/harness.rs
```diff
@@ -12,43 +12,405 @@ use crate::{
...
+    // Load environment
+    let mut env = gadget_config::load(context_config)
+        .map_err(|e| Error::Setup(e.to_string()))
+        .map_err(|err| RunnerError::Other(err.to_string()))?;
+
+    // Always set test mode, dont require callers to set env vars
+    env.test_mode = true;
+    Ok(env)
+}
+
```
Tjemmmic, line 413: We may want a way to generate for new accounts that aren't one of these

crates/testing-utils/tangle/src/tests.rs
```diff
@@ -109,32 +108,6 @@ async fn test_events_client() -> Result<(), Error> {
...
-        "Operator ID should be valid ECDSA public key"
-    );
-
-    // Test blueprint ID retrieval
-    let blueprint_id = harness.client().blueprint_id().await.unwrap();
-    assert!(blueprint_id > 0, "Blueprint ID should be positive");
-
-    Ok(())
-}
-
```
Tjemmmic, line 137: Why was this test removed?

Tjemmmic (approved):

crates/testing-utils/tangle/src/tests.rs
tbraun96: It runs the same code path as several other tests (i.e., redundant)

crates/testing-utils/tangle/src/harness.rs
tbraun96: Long-term, yes. For now, the chainspecs endow only several accounts.

drewstone: I'm with Donovan here. These seem like things we should have now not later.

drewstone: The tester can always endow whatever balances they want. They should be able to run with more than 5 accounts.

PR #629, opened 2025-01-31
feat: testing improvement

drewstone:
Goal with this feature is to get rid of the need to run the runner inside of a thread and instead have this handled in the backend.


drewstone:
@Tjemmmic this might also be good to review

Tjemmmic (approved):

PR #642, opened 2025-02-05
cleanup crate features & update `tangle-subxt`

Serial-ATA:
# Overview
- Use `dep:` syntax more
- Stop explicitly depending on `subxt` and `subxt-core`, use the ones from `tangle-subxt` instead
- Updated `tangle-subxt`
- Added a `PairSigner` type to `gadget-crypto-tangle-pair-signer`, as `subxt-core` has removed its `substrate-compat` feature


drewstone:
The incredible squaring test is failing @Serial-ATA

drewstone:
Realized its broke on main too, CC @Tjemmmic any idea why?

PR #643, opened 2025-02-06
fix: blueprint examples

Tjemmmic:
# Overview

Fixes the blueprint example tests and adds what was missing


drewstone:
Need to build Example blueprint in pipeline to fix clippy at least.

PR #645, opened 2025-02-06
feat(cli): eigenlayer deployment

Tjemmmic:
# Overview

Adds the ability to deploy an Eigenlayer AVS through cargo-tangle


cli/src/deploy/eigenlayer.rs
```diff
@@ -1 +1,278 @@
...
+            .arg("--rpc-url")
+            .arg(&opts.rpc_url)
+            .arg("--private-key")
+            .arg(&private_key);
+
+        // If the contract has constructor args, get them from the user or use provided args
+        if let Some(args) =
+            get_constructor_args(&contract_json, &contract_name, &opts.constructor_args)
+        {
+            if !args.is_empty() {
```
drewstone, line 224: if the args are empty can we prompt the user to provide each arg one at a time?

drewstone (approved):

cli/src/deploy/eigenlayer.rs
Tjemmmic: Yes, it does exactly that

drewstone:
Need to add tutorial to the docs, great work!

PR #666, opened 2025-02-18
feat: Blueprint protocol handshakes

drewstone:


Serial-ATA:
mermaid diagrams are broken by the way

PR #674, opened 2025-02-21
Update to tangle main services

drewstone:


drewstone:
@Serial-ATA check this out, possibly we can just get the update to main knocked out too. This way we get update tangle-subxt properly against main.

PR #679, opened 2025-02-24
Improve appearance and dev experience

drewstone:
Please verify it runs for you @Tjemmmic


.cursor/rules/general.mdc
Tjemmmic, line 1: Did you mean to commit this?

PR #709, opened 2025-03-10
fix: misc fixes

shekohex:


shekohex:
@Serial-ATA shall we close this one?

crates/macros/tests/debug_job/fail/argument_not_extractor.stderr
```diff
@@ -1,4 +1,4 @@
-error[E0277]: the trait bound `bool: FromJobCall<(), blueprint_core::extract::private::ViaParts>` is not satisfied
+error[E0277]: the trait bound `bool: FromJobCall<(), blueprint_sdk::blueprint_core::extract::private::ViaParts>` is not satisfied
```
Serial-ATA, line 1: can you look into why this happens? this isn't a valid path. is it just taking the current crate name for some reason?

not super important, though. it's a path to a private type anyway.

shekohex: this happens because we wrote the tests anyway. also see how we export everything from the blueprint core into our sdk. I guess we wrap it into a `mod core { pub use blueprint_core::*; }` then we can do `pub use self::core::*;` in the sdk.

Serial-ATA: it'd be fine to have `blueprint_sdk::extract::private::ViaParts`, though. where is it getting the `blueprint_core` from?

Serial-ATA:
making another pr tomorrow to fix a couple of tests. I'll take the macro and txhash changes from here when I do

crates/macros/tests/debug_job/fail/argument_not_extractor.stderr
Serial-ATA: reverted this change for now. I'd rather the test fail than us accept a broken path

drewstone (approved):

PR #739, opened 2025-03-14
feat(testing-utils): support per-node contexts

Serial-ATA:
# Overview

The BLS blueprint, since it has networking in the context, needs to be able to set a context per-node, rather than once when the harness is created.

This changes `MultiNodeTestEnv::start()` to take a context, which will be cloned to each node, and adds `MultiNodeTestEnv::start_with_contexts()`, which takes a `Vec<Ctx>` to apply to the nodes in order.

## Example

```rust
use blueprint_core::extract::Context;
use gadget_tangle_testing_utils::TangleTestHarness;
use tempfile::TempDir;

// This context is node specific. Each node needs its own copy.
#[derive(Clone)]
struct MyContext {
    foo: u64,
}

async fn some_job(Context(_context): Context<MyContext>) {}

// Start up a test with 2 nodes
const N: usize = 2;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let tmp_dir = TempDir::new()?;

    // Initialize the harness with the type, this is an indication that it will need the
    // context later.
    let harness = TangleTestHarness::<MyContext>::setup(tmp_dir).await?;

    let (mut test_env, _service_id, _blueprint_id) = harness.setup_services::<N>(false).await?;

    // Setup the test environment
    test_env.initialize().await?;
    test_env.add_job(some_job).await;

    // Ready to start now, provide the contexts
    let mut contexts = Vec::new();
    let handles = test_env.node_handles().await;
    for (index, handle) in handles.iter().enumerate() {
        let context = MyContext { foo: index as u64 };
        contexts.push(context);
    }

    test_env.start_with_contexts(contexts).await?;
    Ok(())
}
```

### DIFF
--- a/cli/src/deploy/tangle.rs
+++ b/cli/src/deploy/tangle.rs
@@ -186,14 +186,14 @@ fn do_cargo_build(manifest_path: &Path) -> Result<()> {
 
     let stdout_thread = thread::spawn(move || {
         let reader = BufReader::new(stdout);
-        for line in reader.lines().flatten() {
+        for line in reader.lines().map_while(Result::ok) {
             tracing::debug!(target: "build-output", "{}", line);
         }
     });
 
     let stderr_thread = thread::spawn(move || {
         let reader = BufReader::new(stderr);
-        for line in reader.lines().flatten() {
+        for line in reader.lines().map_while(Result::ok) {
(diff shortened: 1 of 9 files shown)


shekohex (approved):
I've tested it with frost blueprint, it works.

PR #799, opened 2025-03-26
feat: add targets for some logs

Serial-ATA:
Makes filtering and debugging much easier


drewstone (approved):

shekohex (approved):
Looks good, it would be nice to add docs for all of these.

Serial-ATA:
> Looks good, it would be nice to add docs for all of these.

#802

PR #804, opened 2025-03-26
feat(blueprint-serde): partial map support

Serial-ATA:
This just converts maps into unnamed `Field::Struct`s. That means that map keys must be strings.

So we don't *really* support maps, this is really just a hack for supporting JSON object conversions.


drewstone (approved):

shekohex (commented):
This looks good, could you please add a test to see if nested maps are Supported?

PR #821, opened 2025-04-01
chore: pin workspace to 1.85

Serial-ATA:
Since we don't rely on `rustdoc-types` for metadata anymore, we can pin to the latest stable. All crates have an explicit minimum `rust-version` of 1.85, and the toolchain was downgraded to 1.85.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly"
+channel = "1.85" # Update the workspace rust-version as well
 components = ["rustfmt", "clippy", "rust-src"]
 profile = "minimal"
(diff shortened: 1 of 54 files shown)


drewstone (approved):

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly"
+channel = "1.85" # Update the workspace rust-version as well
```
Tjemmmic, line 2:
```suggestion
channel = "1.85"
```

Tjemmmic (approved):

rust-toolchain.toml
Serial-ATA: Those comments were intentional, if you bump either toolchain version you should bump the workspace `rust-version`

PR #857, opened 2025-04-09
chore: cargo update

Serial-ATA:
`crossbeam-channel` 0.5.14 was yanked. Need to bump it to publish.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4864,9 +4864,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.14"
+version = "0.5.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


shekohex:
I always wonder why we need this in the first place? every single blueprint who want to use networking will have to use this crate! why not just use the `std` channels here?

Serial-ATA:
@shekohex it's not really necessary, but in our case it doesn't matter anyway since it's already in the dependency tree from `libp2p`.

shekohex:
> @shekohex it's not really necessary, but in our case it doesn't matter anyway since it's already in the dependency tree from `libp2p`.

Sure, but I'm talking about the DX here, when I was working on the FROST Blueprint I had to pull this crate (as a direct dependency) just for Networking. Which is not a huge deal for me because I picked the correct version of that crate. It would be nice if we added this somewhere in the docs.

shekohex (approved):

Serial-ATA:
Ah, yeah. We should just re-export crossbeam stuff probably.

PR #905, opened 2025-04-21
refactor(core)!: remove `IntoJobId` trait

Serial-ATA:
The trait was unnecessary, just use `Into<JobId>`.


shekohex:
Can we implement `Into<JobId> for Foo` where I own `Foo`? sometimes clippy complains about this.

Serial-ATA:
Expand on that? Not sure what you mean @shekohex

drewstone (approved):

PR #1117, opened 2025-08-06
feat: add multi tenancy auth

### LINKED ISSUE #1120: [TASK] Add a way to attach additional metadata (headers) during the Auth proxy verify workflow
## Problem

Currently, the upstream service (the final destination) that handles the request behind the `auth-proxy` does not have a way to identify the actual tenant ID, or in our case, the `user_id`, that is sending requests to it.

## Goal

The upstream service needs access to the tenant ID (`user_id`).

### High-Level Solution

* The upstream service is only accessible via the public internet behind the `auth-proxy`. There is no other way to reach it.
* The only method for obtaining a new access token is by signing a challenge with a private key. The public key corresponding to that private key must be an owner of the service in order to exchange the challenge for an access token from the `auth-proxy`.
* The issuer can request to associate additional headers (metadata in a key-value map) with the access token during its generation.
* These headers will be forwarded as-is to the upstream service by the `auth-proxy` whenever it receives a request with a valid access token.
* In this way, the issuer can embed a tenant ID (`user_id`) during token generation, which will then be passed along to the upstream service.

### Implementation Details

* [ ] Add a new field to [types.rs](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/types.rs#L209-L223) to accept additional headers (`additional_headers: BTreeMap<String, String>`), similar to `http::HeaderMap`.
* [ ] This allows access to the headers within the `auth_verify` HTTP handler: [proxy.rs](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/proxy.rs#L136).
* [ ] Update the database models, specifically [`ApiTokenModel`](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/models.rs#L13-L34), to store `additional_headers`.
* [ ] Update [`GeneratedApiToken`](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/api_tokens.rs#L19-L30) to carry this value during the [`generate_token_with_expiration`](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/api_tokens.rs#L75C12-L105) function call.
* [ ] The headers will now be accessible here: [proxy.rs](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/proxy.rs#L176-L177), and stored alongside the generated token in the database.
* [ ] Later, when an HTTP request is received and needs to be sent to the upstream service, after validation, the additional headers must be included in the HTTP request within the `reverse_proxy` handler: [proxy.rs](https://github.com/tangle-network/blueprint/blob/8a285d68ef9a9c45927967583f3cda0972d53583/crates/auth/src/proxy.rs#L253-L261).

## Recap

This approach resolves the issue of the upstream service not having access to the tenant ID (`user_id`) and provides flexibility for future use cases.

## ETA and Timeline

Should be done and tested in less that 4 working hours, with the integration the ai web app.

---

Adopted from 501

drewstone:


drewstone:
@claude review and search for bugs/security issues

shekohex:
Fixes #1120

crates/auth/src/models.rs
```diff
@@ -309,4 +330,66 @@ mod tests {
         assert_eq!(found_token.expires_at, token.expires_at);
         assert_eq!(found_token.is_enabled, token.is_enabled);
     }
+
+    #[test]
+    fn token_with_headers() {
```
shekohex, line 335: Since HTTP headers are case-insensitive, we should lowercase all headers that come from the request and store them as they are (https://stackoverflow.com/a/5259004). This is just my opinion. You should also add a test that tests in this kind of situation: where you add two headers, one with capital cases and the other with lowercase. But at the end of the day, you should only have one, and the one that is added later should override the old one instead of having two different headers.

shekohex (changes requested):
Reviewing more.

shekohex (changes requested):
Left more comments. The overall implementation looks good, but needs to improve the docs on what are all these different tokens and api keys, add examples with typescript to explain the workflow for the integration in the frontend.

drewstone:
@claude review please, meticulously

shekohex (approved):

PR #1192, opened 2025-10-16
feat(blueprint-runner/eigenlayer): update eigen sdk

danielbui12:


crates/manager/src/executor/eigen_event_handler.rs
```diff
@@ -0,0 +1,130 @@
+use crate::config::BlueprintManagerContext;
```
drewstone: This entire file seems wrong. There is nothing exactly _unique_ about Eigenlayer events. Eigenlayer is a protocol on EVM that uses Solidity smart contracts or otherwise EVM compatible languages to emit events. The EVM event handling utilities should be sufficient for anything here.

crates/testing-utils/eigenlayer/src/env.rs
```diff
@@ -11,6 +11,7 @@ use eigensdk::utils::slashing::middleware::registry_coordinator::RegistryCoordin
 use url::Url;
 
 // ================= Core Eigenlayer Deployment Addresses =================
+// TODO(daniel): update these addresses to the new addresses of Tangle testnet
```
drewstone: These have nothing to do with Tangle testnet, we care only about anvil local testnet with forked state, the live testnet eigenlayer uses, and mainnet where eigenlayer is deployed. None of these are Tangle.

PR #1307, opened 2026-02-26
feat(extractors): implement OptionalFromJobCallParts for job extractors

drewstone:
## Summary
Allows `Option<Caller>`, `Option<ServiceId>`, `Option<CallId>`, `Option<JobIndex>`, `Option<JobInputs>` in job handlers. Returns `None` instead of hard-rejecting when metadata is missing.

## Motivation
Each producer (X402Producer, TangleProducer, CronJob) sets different metadata keys on JobCall. Without Optional extraction, a job handler written for on-chain triggers hard-fails when triggered via x402, and vice versa. This is essential for mixed-trigger blueprints.

## Changes
- `crates/evm-extra/src/extract/job.rs` — OptionalFromJobCallParts for ServiceId, CallId, JobIndex, Caller, JobInputs
- `crates/tangle-extra/src/extract/mod.rs` — OptionalFromJobCallParts for all tangle-specific extractors
- Tests for both present and missing metadata scenarios

### DIFF
--- a/crates/evm-extra/src/extract/job.rs
+++ b/crates/evm-extra/src/extract/job.rs
@@ -6,7 +6,7 @@
 use alloy_primitives::Address;
 use blueprint_core::{
     __composite_rejection as composite_rejection, __define_rejection as define_rejection,
-    FromJobCallParts, job::call::Parts as JobCallParts,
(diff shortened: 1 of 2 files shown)


drewstone:
@claude review

PR #1341, opened 2026-03-22
fix: publish crates in topological dependency order

drewstone:
## Summary

batch-publish.sh was publishing crates in arbitrary order, causing failures when a crate dependency had not been published to crates.io yet (e.g. blueprint-crypto-hashing before blueprint-std).

Now sorts by internal dependency count using cargo metadata (leaves first, root crates like blueprint-sdk last). Adds retry with 30s backoff for transient registry index lag. Continues on individual failures instead of aborting the entire batch.

## Change Class

- Selected class: Class B
- Why this class: CI script only, no runtime code changes.

## Behavior Contract

- Crates publish in topological order (leaf deps first)
- Failed crate retried once after 30s
- Individual failures no longer abort the entire batch
- Burst delay increased from 2s to 5s (registry index propagation)

## Risk And Scope

- Scope: 1 file, CI script
- Risk: Low — only affects publish ordering

## Verification

\`\`\`bash
bash .github/scripts/batch-publish.sh test-input.json
\`\`\`

## Harness Evidence

Script change only.

## Checklist

- [x] Script tested locally
- [x] No code changes

### DIFF
--- a/.github/scripts/batch-publish.sh
+++ b/.github/scripts/batch-publish.sh
@@ -1,13 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
-# This script publishes Rust crates while respecting crates.io's rate limit.
+# This script publishes Rust crates in topological dependency order while
+# respecting crates.io's rate limit.
+#
 # Rate limit: burst=30 (immediate), then 1 request per minute
 # See: https://github.com/rust-lang/crates.io/blob/master/src/middleware/app.rs
 #
-# Strategy:
-# - Publish first 30 crates immediately (using the burst allowance)
-# - Wait 60 seconds between each subsequent crate (1 per minute rate)
+# Key: crates are sorted so leaf dependencies publish first. Without this,
+# a crate that depends on blueprint-std will fail if blueprint-std hasn't
+# been published yet, because cargo publish --no-verify still checks that
+# all dependencies resolve on the registry.
 
 RELEASE_JSON="${1:-}"
 BURST_SIZE=30       # Can publish 30 immediately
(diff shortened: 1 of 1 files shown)


drewstone:
## ❌ PR Review w/ codex, claude

| | |
|---|---|
| **Recommendation** | Needs Work |
| **Findings** | 5 total — 🔴 1 high, 🟠 3 medium, 🟡 1 low |
| **Ensemble** | 2 reviewers × 2 tracks |
| **Files reviewed** | 1 files changed |
| **Validator** | claude |
| **Provenance** | [individual reviewer outputs](https://gist.github.com/drewstone/ae1522fb237c803c507e9e85280048a1) |

---

### 🔴 HIGH (1)

- **Sorting by dependency count is not a topological sort — will misorder crates in this workspace** [`.github/scripts/batch-publish.sh:80`](https://github.com/tangle-network/blueprint/blob/d9dd946a0c9d49c2799f0cc9ad02f1bb0526a611/.github/scripts/batch-publish.sh#L80)
  > The Python snippet sorts packages by direct internal dependency count ascending, then alphabetically. This is not a topological sort. A topological sort guarantees that for every edge A→B, B appears before A. Sorting by count makes no such guarantee.

Concrete failure in this workspace: blueprint-client-tangle has 5 normal internal deps and depends on blueprint-crypto (crates/clients/tangle/Cargo.toml line 21). blueprint-crypto has 7 internal deps (the crypto subcrates). The sort places client-t…

### 🟠 MEDIUM (3)

- **Dev-dependencies counted in sort key inflate counts and distort ordering** [`.github/scripts/batch-publish.sh:88`](https://github.com/tangle-network/blueprint/blob/d9dd946a0c9d49c2799f0cc9ad02f1bb0526a611/.github/scripts/batch-publish.sh#L88)
  > The Python snippet iterates all entries in pkg['dependencies'] without filtering by kind. cargo metadata includes dev-dependencies (kind='dev') and build-dependencies (kind='build') in this array. Dev-deps don't need to be published first — cargo publish only requires normal dependencies to resolve on the registry. Including dev-deps inflates counts and further distorts ordering relative to what cargo publish actually requires.

- **Continue-on-failure cascades through dependency chains** [`.github/scripts/batch-publish.sh:171`](https://github.com/tangle-network/blueprint/blob/d9dd946a0c9d49c2799f0cc9ad02f1bb0526a611/.github/scripts/batch-publish.sh#L171)
  > After a package fails twice, the script records it and continues. But every downstream crate depending on the failed one will also fail (cargo publish checks that all dependencies resolve on the registry). Each downstream failure triggers a 30s wait + retry + inter-publish delay. For a foundational crate like blueprint-core failing, 10+ dependent crates each burn ~35+ seconds on doomed retries, producing noisy output that obscures the root cause. The loop should skip descendants of failed packag…

- **Suggested recovery path is not idempotent after partial publish** [`.github/scripts/batch-publish.sh:200`](https://github.com/tangle-network/blueprint/blob/d9dd946a0c9d49c2799f0cc9ad02f1bb0526a611/.github/scripts/batch-publish.sh#L200)
  > The final report tells operators to 'Re-run this script' after partial success, but on rerun, crates that succeeded in the first pass will fail with crates.io's 'version already exists' error, be retried, and be added to failed_packages. The recommended recovery path will continue to exit non-zero even if all previously-missing crates publish successfully.

### 🟡 LOW (1)

- **Workspace member detection uses fragile substring matching (harmless with --no-deps)** [`.github/scripts/batch-publish.sh:76`](https://github.com/tangle-network/blueprint/blob/d9dd946a0c9d49c2799f0cc9ad02f1bb0526a611/.github/scripts/batch-publish.sh#L76)
  > Line 76 uses pkg['name'] in m as a fallback check, which is a substring match against workspace_members entries. A package named 'blueprint' would match an entry for 'blueprint-sdk'. With --no-deps, all packages in meta['packages'] are already workspace members, so the matching is currently redundant and harmless. But if --no-deps behavior changes or the script is adapted, substring matching could cause false positives.


---

### 🎯 What would get this approved

- [ ] Replace the dep-count sort (lines 80-95 of the Python snippet) with a real topological sort (Kahn's algorithm or DFS post-order). The dependency edges are already available — this is ~15-20 lines of Python, replacing ~10 existing lines. The sort must only count normal dependencies (filter out kind='dev' and kind='build').
- [ ] Track failed packages in the publish loop and skip any crate whose normal dependencies include an already-failed package, logging 'Skipping X because dependency Y failed' instead of attempting a doomed publish+retry cycle.
- [ ] Handle 'version already exists' as a success (exit 0 equivalent) in the publish loop so that reruns are idempotent. cargo publish returns a recognizable error message for this case.

---
<sub>pr-reviewer v0.5.0 · review #1 · 2026-03-22T08:17:48.097206+00:00</sub>

PR #1457, opened 2026-06-14
fix(release): batch-publish waits on the sparse index, not the web API

drewstone:
## Summary

- The crates.io batch publish stalled at its first no-progress pass, leaving ~55 crates (including `blueprint-sdk`) unpublished after a release. Workspace versions are consistent — this is a propagation-timing bug in `batch-publish.sh`, not a version bug.
- `cargo publish` resolves a crate's dependencies against the **sparse index** (`index.crates.io`), which propagates slower than the crates.io **web API**. The old `version_is_live` polled the web API, declared a just-published dependency "live", and moved on while cargo still couldn't see it — so every dependent failed `failed to select a version for X = ^alpha.N` and the retry loop bailed on its no-progress guard.
- Fix: check **and wait on the sparse index** (the same source cargo reads), in both the main pass and the retry loop.

## Change Class

- `Class A` (docs/tooling only)
- Selected class: Class A
- Why this class: a single CI release script under `.github/`; no library/runtime code.

## Behavior Contract

- Current behavior: the publisher polls the web API, moves on before the sparse index has the version, dependents can't resolve it, the run stalls and exits 1 with dozens of crates unpublished.
- Intended behavior: after publishing a crate, poll the sparse index until the exact version is visible (up to 5 min, exits early), so a dependency is provably cargo-resolvable before its dependents publish next in topo order.
- Invariants that must hold: publish order stays topological (leaves first); a crate already in the sparse index is skipped; the run only reports success when every version is sparse-index-live.
- Failure mode choice: fail-closed — a version that never appears in the sparse index within the timeout is recorded as failed rather than silently assumed published.

## Risk And Scope

- Security impact: none.
- Compatibility impact: none — only the publish script's readiness check changes; no crate contents or versions change.
- Migration notes: none.
- Rollback plan: revert the one script.

## Verification

```bash
bash -n .github/scripts/batch-publish.sh   # syntax
# sparse_index_path + version_is_live correctly detect a live version:
curl -fsSL https://index.crates.io/bl/ue/blueprint-networking | grep '"vers":"0.2.0-alpha.8"'
```

Outcome: syntax clean; the new sparse-index path (`bl/ue/blueprint-networking`) and check correctly report `blueprint-networking 0.2.0-alpha.8` as live. The next `publish-crates.yml` run is the end-to-end proof (it should now converge the remaining crates).

## Harness Evidence

- Reproducer added/updated: the stalled publish run is the reproducer; the sparse-index wait removes the cascade that caused it.
- Negative-path coverage added: `wait_until_live` bounds the wait and records a genuine miss as failed (no false success).
- Key assertions that prove the fix: the sparse-index check resolves a known-live version; dependents now publish because their deps are sparse-index-visible first.

## Checklist

- [x] I evaluated compatibility/migration impact explicitly.
- [x] I validated script syntax and the sparse-index check against a live version.
- [x] No crate contents, versions, or runtime behavior are affected.

### DIFF
--- a/.github/scripts/batch-publish.sh
+++ b/.github/scripts/batch-publish.sh
@@ -149,6 +149,39 @@ echo ""
 # Track failures for retry
 declare -a failed_packages=()
 
+# crates.io SPARSE-INDEX path for a crate name. `cargo publish` resolves a
+# crate's dependencies against the sparse index (index.crates.io), which
+# propagates SLOWER than the web API. The old check polled the web API, so the
+# script would declare a dep "live" and move on while cargo still couldn't see
+# it — every dependent then failed "failed to select a version" and the run
+# stalled at the first no-progress pass. We must wait on the sparse index.
+sparse_index_path() {
+    local name="$1" len=${#name}
+    if   ((len == 1)); then echo "1/$name"
+    elif ((len == 2)); then echo "2/$name"
+    elif ((len == 3)); then echo "3/${name:0:1}/$name"
+    else echo "${name:0:2}/${name:2:2}/$name"
+    fi
+}
+
+# Is this exact version visible in the sparse index (i.e. cargo-resolvable)?
+version_is_live() {
+    local name="$1" version="$2"
+    curl -fsSL --max-time 10 "https://index.crates.io/$(sparse_index_path "$name")" 2>/dev/null \
+        | grep -q "\"vers\":\"$version\""
+}
+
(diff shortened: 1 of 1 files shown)


tangletools (approved):
## ✅ Auto-approved PR — `692d18c6`

Blanket team auto-approval is enabled for this reviewer service.

<sub>tangletools · auto-approval · reason: `blanket_auto_approve` · 2026-06-14T11:07:40Z</sub>

PR #1467, opened 2026-07-07
feat: adopt tnt-core 0.19 bindings (draft — blocked on publishing tnt-core-bindings 0.19.0)

drewstone:
## Summary
Adopt tnt-core 0.19 bindings in the operator manager + CLI (`binaryUri` is event-sourced now, not a view-struct field) and fix the multi-shot review blockers: scope the binary-URI `getLogs` query to the Tangle contract, bump `crossbeam-epoch` for RUSTSEC-2026-0204, and make version listing tolerate a missing publish log.

## Change Class
- Selected class: Class C — behavior-affecting across multiple crates (`blueprint-manager` + `cargo-tangle`), including a security fix and a dependency security bump.
- Why this class: changes the manager's on-chain read path (binaryUri sourced from the `BinaryVersionPublished` event, address-scoped) and the CLI's version listing; not a pure refactor, not config/docs-only.

## Behavior Contract
- Current behavior: manager/CLI read `binaryUri` from the `getBinaryVersion` view struct (removed in tnt-core 0.19); the URI resolver's `getLogs` filter matched `BinaryVersionPublished` from any contract.
- Intended behavior: `binaryUri` is sourced from the `BinaryVersionPublished` event scoped to the configured Tangle contract address; the read-only version listing blanks an unresolvable URI instead of aborting.
- Invariants that must hold: the manager only accepts a `binaryUri` emitted by the configured Tangle contract for the exact (blueprintId, versionId); `download_and_verify` still verifies the sha256; the upgrade path still fails hard when the URI is missing.
- Failure mode choice (`fail-closed` or `fail-open`) and rationale: fail-closed for the manager/upgrade path (missing/unscoped URI -> error, never a silent wrong-source download); fail-open for the read-only listing (blank row) so one pruned log does not hide every version.

## Risk And Scope
- Security impact: closes a binary-URI spoofing vector — any contract could previously satisfy the unscoped `getLogs` filter and feed an attacker-controlled download URL into `download_and_verify`; also bumps `crossbeam-epoch` 0.9.18 -> 0.9.20 (RUSTSEC-2026-0204).
- Compatibility impact (APIs, metadata format, configs): consumes `tnt-core-bindings` 0.19.0 (published); the `BinaryVersion` struct no longer carries `binaryUri`; no public API of this repo changes.
- Migration notes (if any): none — operators pick up the new binding on the next build; the on-chain event already carries `binaryUri`.
- Rollback plan: revert the branch; the 0.18 bindings + view-based read still resolve against 0.18 chains.

## Verification
- Reproducer added/updated: the security fix mirrors the already-correct CLI path (`.address(ctx.tangle_contract)`); covered by build + existing upgrade tests.
- Negative-path coverage added: version listing tolerates a missing publish log (`Ok(None)` -> blank row); the upgrade path still errors on a missing URI via `require_binary_uri`.
- Key assertions that prove the fix, with commands:
  - `cargo build -p blueprint-manager -p blueprint-qos -p cargo-tangle` -> 0 errors
  - `cargo clippy -p blueprint-manager -p cargo-tangle` -> 0 warnings on touched files

## Harness Evidence
- `cargo build -p blueprint-manager -p blueprint-qos -p cargo-tangle` -> 0 errors.
- `cargo +nightly fmt -- --check` -> clean.
- Cargo.lock diff minimized to `tnt-core-bindings` 0.19.0 + `crossbeam-epoch` 0.9.20.

## Checklist
- [x] Behavior contract filled and accurate.
- [x] Security and compatibility impact assessed.
- [x] Verification and harness evidence attached.
- [x] Rollback plan stated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -169,16 +169,16 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b44937ce84d2cbf1ee4010667bd9214bb7db134a91dd9ffa6b1b8b1d6b030449"
+checksum = "83447eeb17816e172f1dfc0db1f9dc0b7c5d069bd1f7cecbecceb382bf931015"
 dependencies = [
- "alloy-eips 2.1.0",
+ "alloy-eips 2.0.5",
  "alloy-primitives",
  "alloy-rlp",
- "alloy-serde 2.1.0",
+ "alloy-serde 2.0.5",
  "alloy-trie",
- "alloy-tx-macros 2.1.0",
+ "alloy-tx-macros 2.0.5",
  "auto_impl",
  "borsh",
  "c-kzg",
@@ -210,15 +210,15 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus-any"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7bb0bdd61ddff726026676450e7e6a52c68aa39d98f2d60d05ad7caaea5b8100"
+checksum = "5406343e306856dc2be762700e98a16904de45dee14a07f233e742ce68daff2f"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-eips 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-eips 2.0.5",
  "alloy-primitives",
  "alloy-rlp",
- "alloy-serde 2.1.0",
+ "alloy-serde 2.0.5",
  "serde",
 ]
 
@@ -248,9 +248,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-core"
-version = "1.6.0"
+version = "1.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "62ddde5968de6044d67af107ad835bc0069a7ca245870b94c5958a7d8712b184"
+checksum = "23e8604b0c092fabc80d075ede181c9b9e596249c70b99253082d7e689836529"
 dependencies = [
  "alloy-dyn-abi",
  "alloy-json-abi",
@@ -273,7 +273,7 @@ dependencies = [
  "itoa",
  "serde",
  "serde_json",
- "winnow 1.0.3",
+ "winnow 1.0.1",
 ]
 
 [[package]]
@@ -317,30 +317,14 @@ dependencies = [
 
 [[package]]
 name = "alloy-eip7928"
-version = "0.3.7"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b827a6d7784fe3eb3489d40699407a4cdcce74271421a01bdffe60cf573bb16"
+checksum = "f8222b1d88f9a6d03be84b0f5e76bb60cd83991b43ad8ab6477f0e4a7809b98d"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
  "borsh",
- "once_cell",
  "serde",
- "thiserror 2.0.18",
-]
-
-[[package]]
-name = "alloy-eip7928"
-version = "0.4.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b3b12337f74cbfa451cb04dac173974814a6ff463079e1793aa09600ba8813ab"
-dependencies = [
- "alloy-primitives",
- "alloy-rlp",
- "borsh",
- "once_cell",
- "serde",
- "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -352,7 +336,7 @@ dependencies = [
  "alloy-eip2124",
  "alloy-eip2930",
  "alloy-eip7702",
- "alloy-eip7928 0.3.7",
+ "alloy-eip7928",
  "alloy-primitives",
  "alloy-rlp",
  "alloy-serde 1.8.3",
@@ -368,17 +352,17 @@ dependencies = [
 
 [[package]]
 name = "alloy-eips"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "154b0f566ebbfc256b63c8d643c6a299f40a6fcd50a9d756c1d191487dd06483"
+checksum = "0dca4c89ace90684b4b77366d00631ed498c9af962079af2a5dbc593a0618a77"
 dependencies = [
  "alloy-eip2124",
  "alloy-eip2930",
  "alloy-eip7702",
- "alloy-eip7928 0.4.5",
+ "alloy-eip7928",
  "alloy-primitives",
  "alloy-rlp",
- "alloy-serde 2.1.0",
+ "alloy-serde 2.0.5",
  "auto_impl",
  "borsh",
  "c-kzg",
@@ -424,7 +408,7 @@ checksum = "422d110f1c40f1f8d0e5562b0b649c35f345fccb7093d9f02729943dcd1eef71"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
- "http 1.4.2",
+ "http 1.4.0",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
@@ -433,13 +417,13 @@ dependencies = [
 
 [[package]]
 name = "alloy-json-rpc"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2e92108767c8c95b5e521570e039e374e65198c4179a98d200412c083d6c26d5"
+checksum = "ec0a82e56b1843bce483942d54fcadea92e676f1bde162e93c7d3b621fabc4e1"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
- "http 1.4.2",
+ "http 1.4.0",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
@@ -474,20 +458,20 @@ dependencies = [
 
 [[package]]
 name = "alloy-network"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c4546c9fae2861d4159ee3b25c44ab3edb90f21b849e86cf2086cacb1d56d8ac"
+checksum = "a7db7b095b0b1db1d18ce7e91dcd2e82007f2d52bfb8125e6b64633a74a06bc3"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-consensus-any 2.1.0",
- "alloy-eips 2.1.0",
- "alloy-json-rpc 2.1.0",
- "alloy-network-primitives 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-consensus-any 2.0.5",
+ "alloy-eips 2.0.5",
+ "alloy-json-rpc 2.0.5",
+ "alloy-network-primitives 2.0.5",
  "alloy-primitives",
- "alloy-rpc-types-any 2.1.0",
- "alloy-rpc-types-eth 2.1.0",
- "alloy-serde 2.1.0",
- "alloy-signer 2.1.0",
+ "alloy-rpc-types-any 2.0.5",
+ "alloy-rpc-types-eth 2.0.5",
+ "alloy-serde 2.0.5",
+ "alloy-signer 2.0.5",
  "alloy-sol-types",
  "async-trait",
  "auto_impl",
@@ -513,14 +497,14 @@ dependencies = [
 
 [[package]]
 name = "alloy-network-primitives"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca2c325d5934445209c8d22fc67d27e2cf9fb79054b8154d9e522d6a4ee3a4f7"
+checksum = "cd28d9bfd11729037d194f2b1d43db8642eb3f342032691f4ca96bb745479c3c"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-eips 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-eips 2.0.5",
  "alloy-primitives",
- "alloy-serde 2.1.0",
+ "alloy-serde 2.0.5",
  "serde",
 ]
 
@@ -536,7 +520,7 @@ dependencies = [
  "const-hex",
  "derive_more",
  "foldhash 0.2.0",
- "hashbrown 0.17.1",
+ "hashbrown 0.17.0",
  "indexmap 2.14.0",
  "itoa",
  "k256",
@@ -546,7 +530,7 @@ dependencies = [
  "rand 0.9.4",
  "rapidhash",
  "ruint",
- "rustc-hash 2.1.3",
+ "rustc-hash 2.1.2",
  "secp256k1 0.31.1",
  "serde",
  "sha3 0.11.0",
@@ -588,7 +572,7 @@ dependencies = [
  "lru",
  "parking_lot 0.12.5",
  "pin-project",
- "reqwest 0.13.4",
+ "reqwest 0.13.2",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
@@ -622,9 +606,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rlp"
-version = "0.3.16"
+version = "0.3.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24671b1f62edcf0f9b62994c7bf72cd621a04a4b99f5020ece1a647b40e2f103"
+checksum = "dc90b1e703d3c03f4ff7f48e82dd0bc1c8211ab7d079cd836a06fcfeb06651cb"
 dependencies = [
  "alloy-rlp-derive",
  "arrayvec",
@@ -633,13 +617,13 @@ dependencies = [
 
 [[package]]
 name = "alloy-rlp-derive"
-version = "0.3.16"
+version = "0.3.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d4311c03125e8a18296504560b9de3d75ecbd0dcda7f71e6cf2a196d57e6fba"
+checksum = "f36834a5c0a2fa56e171bf256c34d70fca07d0c0031583edea1c4946b7889c9e"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.118",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -657,7 +641,7 @@ dependencies = [
  "alloy-transport-ws",
  "futures",
  "pin-project",
- "reqwest 0.13.4",
+ "reqwest 0.13.2",
  "serde",
  "serde_json",
  "tokio",
@@ -670,17 +654,17 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-client"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f8d765656f02f993fa0565abeb3554ccd6a0d72ea44ce0558b983136639bd6b"
+checksum = "24f461f091dc8f657e73b5dea18fd63d5c7049720cd252f1eade4a7ebed6a7e1"
 dependencies = [
- "alloy-json-rpc 2.1.0",
+ "alloy-json-rpc 2.0.5",
  "alloy-primitives",
- "alloy-transport 2.1.0",
- "alloy-transport-http 2.1.0",
+ "alloy-transport 2.0.5",
+ "alloy-transport-http 2.0.5",
  "futures",
  "pin-project",
- "reqwest 0.13.4",
+ "reqwest 0.13.2",
  "serde",
  "serde_json",
  "tokio",
@@ -733,15 +717,15 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-any"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b06bfe79149dd53de5b196aa3c96db0500b7cf0ed72dbd1a31bac7660bc7cc65"
+checksum = "0a6561ed4759c974d9c144500a59e3fb8c1d87327a12900d5ce455c0cae6dcb6"
 dependencies = [
- "alloy-consensus-any 2.1.0",
- "alloy-network-primitives 2.1.0",
+ "alloy-consensus-any 2.0.5",
+ "alloy-network-primitives 2.0.5",
  "alloy-primitives",
- "alloy-rpc-types-eth 2.1.0",
- "alloy-serde 2.1.0",
+ "alloy-rpc-types-eth 2.0.5",
+ "alloy-serde 2.0.5",
  "serde",
  "serde_json",
 ]
@@ -798,17 +782,17 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-eth"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b41d1a11f58b456c199e04591befc64cb236fa2574a7275a07b98b173727bd39"
+checksum = "175a2a5b6017d7f61b5e4b800d21215fe8e94fe729d00828e13bb6d93dcf3492"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-consensus-any 2.1.0",
- "alloy-eips 2.1.0",
- "alloy-network-primitives 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-consensus-any 2.0.5",
+ "alloy-eips 2.0.5",
+ "alloy-network-primitives 2.0.5",
  "alloy-primitives",
  "alloy-rlp",
- "alloy-serde 2.1.0",
+ "alloy-serde 2.0.5",
  "alloy-sol-types",
  "itertools 0.14.0",
  "serde",
@@ -856,9 +840,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-serde"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bcf028ac8bcb161ad7c104243e710cece8e1a794f65ee6c35c4c4d693c8e80ee"
+checksum = "cc21a8772af7d78bba286726aa245bd2ff81cd9abe230afea2e91578996831c9"
 dependencies = [
  "alloy-primitives",
  "serde",
@@ -884,9 +868,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-signer"
-version = "2.1.0"
+version = "2.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a5f392f2f56b2417ea6b74e1e116c6f35df5ab5fce4dc422e277d72ee18ff7b"
+checksum = "8ffbce94c50dd9d4d1f83e044c5595bbd3ada981bd3057ce28b3a5470e77385d"
 dependencies = [
  "alloy-primitives",
  "async-trait",
@@ -918,14 +902,14 @@ dependencies = [
 
 [[package]]
 name = "alloy-signer-aws"
-version = "2.1.0"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0dc378541fe938c036f0000029a759c921d2c9bfd9c2b7a500a93787a2d65e69"
+checksum = "e7a57d1e72b1f9b11e5e71ebdab0569cb02277a462bbea6793fcaebfcd794ae9"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-network 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-network 2.0.5",
  "alloy-primitives",
- "alloy-signer 2.1.0",
+ "alloy-signer 2.0.5",
  "async-trait",
  "aws-config",
  "aws-sdk-kms",
@@ -937,14 +921,14 @@ dependencies = [
 
 [[package]]
 name = "alloy-signer-gcp"
-version = "2.1.0"
+version = "2.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e368b8f5e157d4114804580c827c3bb2294bc48032475214d1e6d09a71702401"
+checksum = "7ffc2a49bca5b73c6964711b57452f6c36a6bcb7f845ab7e9ad05b5a828d0161"
 dependencies = [
- "alloy-consensus 2.1.0",
- "alloy-network 2.1.0",
+ "alloy-consensus 2.0.5",
+ "alloy-network 2.0.5",
  "alloy-primitives",
- "alloy-signer 2.1.0",
+ "alloy-signer 2.0.5",
  "async-trait",
  "gcloud-sdk",
  "k256",
@@ -1003,7 +987,7 @@ dependencies = [
  "proc-macro-error2",
  "proc-macro2",
  "quote",
- "syn 2.0.118",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -1021,7 +1005,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "sha3 0.11.0",
- "syn 2.0.118",
+ "syn 2.0.117",
  "syn-solidity",
 ]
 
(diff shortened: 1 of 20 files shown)


tangletools (approved):
## ✅ Auto-approved drewstone PR — `35d342d0`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-07T22:33:01Z</sub>

tangletools (commented):
## 🟠 Value Audit — better-approach-exists

| | |
|---|---|
| **Verdict** | better-approach-exists |
| **Concerns** | 4 (1 medium-concern, 1 low, 2 weak-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 131.0s (2 bridge agents) |
| **Total** | 131.0s |

### 💰 Value — better-approach-exists

Necessary tnt-core 0.19 ABI adoption that sources binary URIs from event logs, but it duplicates the existing TangleClient windowed event-scan pattern.

- **What it does:** Adopts the tnt-core 0.19 ABI break: `BinaryVersion` loses its `binaryUri` field, so the manager and CLI now read the URI back from the `BinaryVersionPublished` event using bounded, newest-first `eth_getLogs` scans. It also updates the local `sol!` struct field order and pulls in cascading dependency API changes (prometheus, opentelemetry, libp2p, tonic).
- **Goals it achieves:** Keeps binary fetching, version listing, and upgrade flows working after the upstream contract stopped returning `binaryUri` from views; makes the workspace build against `tnt-core-bindings 0.19`.
- **Assessment:** Good and coherent. The ABI-stub intermediate state is clearly marked with TODOs, the event scan is bounded and provider-friendly, and the mechanical dependency fallout is handled correctly.
- **Better / existing approach:** The codebase already has an almost identical windowed `eth_getLogs` scan in `TangleClient::get_raw_blueprint_definition_from_event` (`crates/clients/tangle/src/client.rs:822`), and already has `read_u64_topic` in `crates/tangle-extra/src/producer.rs:661`. The manager's `ChainView` already owns a `TangleClient`, so instead of adding two new `u64_topic` helpers and windowed-scan loops, the PR should
- **Model:** opencode/kimi-for-coding/k2p7
- **Bridge attempts:** 1

Necessary ABI-break adoption that correctly event-sources the dropped `binaryUri` field; fully wired into both CLI and manager, with two small consistency gaps against the codebase's own established bounded-scan pattern.

- **Integration:** Fully reachable on both paths. CLI: `list_versions`/`get_version`/`get_effective_version` are called from `cli/src/main.rs:2234,2259,2266,2488,2507` and flow through `VersionView` into all three print/json helpers (`upgrade.rs:1213,1261,1402`). Manager: `ChainView::get_binary_version` is called from `crates/manager/src/upgrade/rpc.rs:268,357` and `effective_binary_version` from `watcher.rs:200` an
- **Fit with existing patterns:** Follows the established event-sourcing pattern (`TangleClient::get_raw_blueprint_definition_from_event` at `crates/clients/tangle/src/client.rs:822`) which also walks newest-first in bounded windows for an indexed event. The new code reimplements that loop shape correctly. The only friction: it hardcodes the window and doesn't reuse the codebase's configurable `max_getlogs_range`/`resolve_max_getl
- **Real-world viability:** Works on the happy path (recent version → one or two requests). Two gaps vs the established pattern: (1) the manager's filter omits the contract address scope, (2) both copies hardcode a 10_000 window while the codebase default is a configurable 2_000 — the existing comment at `client.rs:50` notes Base Sepolia rejects spans >2000, so the larger constant may re-trigger the caps the configurable app
- **Bridge attempts:** 1

### 🔎 Heuristic Signals

🟡 **Cruft: todo added** `crates/manager/src/upgrade/abi.rs`
> +// TODO: replace with tnt-core-bindings typed `IBlueprintBinaryVersions` once published.

### 🎯 Usefulness Audit

🟡 **Manager's event filter omits the contract address scope (CLI copy has it)** `[robustness]` ``
> `crates/manager/src/upgrade/chain.rs:154-159` builds the `BinaryVersionPublished` filter with only `event_signature + topic1 + topic2`, missing `.address(self.tangle_address)`. The CLI copy at `cli/src/command/upgrade.rs` correctly includes `.address(ctx.tangle_contract)`. The doubly-indexed topic filter makes a foreign-contract collision very unlikely, and the established sibling `get_raw_blueprint_definition_from_event` (`client.rs:846`) also omits the address but compensates with a keccak int

🟡 **Hardcoded 10_000 window vs the codebase's configurable getlogs range** `[fit]` ``
> Both `chain.rs:139` and `upgrade.rs` use `const GETLOGS_WINDOW: u64 = 10_000`. The codebase already solved this per-chain via `resolve_max_getlogs_range()` (`client.rs:58`, default 2_000, env-overridable) precisely because some RPCs cap spans hard — the comment at `client.rs:50` calls out Base Sepolia rejecting >2000 with -32602. The manager's `ChainView` holds a `TangleClient` that already carries `max_getlogs_range`; aligning to that value (or exposing it) would be consistent and avoid re-trig

### 💰 Value Audit

🟠 **Windowed event-scan and u64 topic encoding are duplicated** `[duplication]` ``
> Two new `u64_topic` helpers and nearly identical bounded `eth_getLogs` loops were added in `crates/manager/src/upgrade/chain.rs:135` and `cli/src/command/upgrade.rs:767`. The existing `TangleClient` already implements the same pattern for blueprint definitions (`crates/clients/tangle/src/client.rs:822`), and `read_u64_topic` already lives in `crates/tangle-extra/src/producer.rs:661`. Consolidating this into a shared helper on `TangleClient` or in `blueprint-evm-extra` would remove the duplicatio

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260707T223700Z</sub>

tangletools:
## ❌ Needs Work — `35d342d0`

**Review health 100/100** · **Reviewer score 0/100** · **Confidence 95/100** · 29 findings (1 critical, 3 high, 6 medium, 19 low)

| | deepseek | kimi-code | aggregate |
|---|---|---|---|
| **Readiness** | 32 | 0 | **0** |
| **Confidence** | 95 | 95 | **95** |
| **Correctness** | 32 | 0 | **0** |
| **Security** | 32 | 0 | **0** |
| **Testing** | 32 | 0 | **0** |
| **Architecture** | 32 | 0 | **0** |

_Reviewer score is advisory once the run is complete and the verdict has no blockers._

> Full multi-shot audit completed 8/8 planned shots over 20 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 20 changed files. Global verifier still owns final merge decision.

### Blocking

🟣 **CRITICAL** Prometheus metric logging code does not compile — [`crates/qos/src/metrics/prometheus/server.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/metrics/prometheus/server.rs#L161)
> [Line 161](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/metrics/prometheus/server.rs#L161) `m.get_counter().get_value()` and [line 164](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/metrics/prometheus/server.rs#L164) `m.get_gauge().get_value()` invoke `MessageFieldExt::get_value()`, which lives in prometheus's private `proto_ext` module and is not re-exported. In prometheus 0.14.0 (unchanged in Cargo.lock), `Metric::get_counter()` returns `&MessageField<Counter>`; `MessageField<Counter>` derefs to `Counter`, and `Counter::value()` returns `f64`. Replace both `.get_value()` calls with `.value()` so the handler compiles.

🔴 **HIGH** Cargo.lock still pins vulnerable crossbeam-epoch 0.9.18 — [`Cargo.lock`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.lock#L5316)
> [Line 5316](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.lock#L5316) records `crossbeam-epoch = "0.9.18"`. This version is affected by RUSTSEC-2026-0204 (invalid pointer dereference in `fmt::Pointer` for `Atomic`/`Shared`). In this workspace `cargo update -p crossbeam-epoch` selects `0.9.20` with no other lock changes, so the fix is trivial and should be included before merge.

🔴 **HIGH** hickory-proto upgraded to vulnerable 0.25.2 — [`Cargo.lock`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.lock#L7570)
> [Line 7570](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.lock#L7570) records `hickory-proto = "0.25.2"`. This version is affected by RUSTSEC-2026-0118 (NSEC3 closest-encloser proof validation enters an unbounded loop / OOM on cross-zone responses) for which no patched 0.25 release exists. The base lock used hickory-proto 0.24.4, which is unaffected. The exposure is forced by libp2p-dns 0.44.0 (libp2p 0.56.0) requiring ^0.25; track an upstream hickory/libp2p bump or pin to an unaffected branch once feasible.

🔴 **HIGH** URI resolver does not scope getLogs to the Tangle contract address — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs#L154-L159)
> The Filter built at [lines 154-159](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs#L154-L159) sets event_signature, topic1, and topic2 but never calls .address(self.tangle_address). Because event selectors and indexed topics are global, any contract can emit a BinaryVersionPublished(blueprintId, versionId, sha256Hash, binaryUri) log that satisfies the filter. The manager then iterates logs newest-first and returns the first decodable URI, so an attacker-controlled log can override the legitimate URI, causing DoS, operator privacy leaks, or trust-root confusion. The CLI equivalent at cli/src/command/upgrade.rs:793 correctly includes .address(ctx.tangle_contract). Fix: add .address(self.tangle_address) to the Filt

### Other

🟠 **MEDIUM** jsonwebtoken v10 defaults validate_exp to false — defense-in-depth reduced — [`Cargo.toml`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.toml)
> jsonwebtoken 9→10.3 changes Validation::new(alg) to default validate_exp=false (was true in v9). Consumers auth/src/oauth/mod.rs:156 and blueprint-remote-providers/src/auth_integration.rs:240 call Validation::new(alg) without explicitly setting validate_exp=true. All affected paths have manual expiry checks that prevent actual token lifetime bypass, so this is not exploitable. However, future code that relies on the library's built-in validation could silently accept expired tokens. Recommend adding explicit validate_exp = true at both call sites for defense-in-depth and to make the intent clear to contributors.

🟠 **MEDIUM** Fresh provider per URI lookup wastes connections — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs)
> Each call builds a new ProviderBuilder and opens a new connection. list_versions invokes this once per version, so N versions create N+1 providers. The manager implementation (crates/manager/src/upgrade/chain.rs) reuses a shared TangleClient. Refactor to accept or cache a shared provider, or add a list-scoped helper that resolves URIs in one pass.

🟠 **MEDIUM** Redundant Provider connections + no retry in binary_uri_from_event — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L777-L780)
> binary_uri_from_event creates a fresh Provider connection every call ([line 777-780](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L777-L780)), independently of call_view which also creates one ([line 1048-1051](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L1048-L1051)). For list_versions this means 2*count HTTP connections per invocation plus count*N eth_getLogs calls where N = chain_blocks/10000. Each eth_getLogs has no retry — any transient RPC error fails the entire operation. Recommend: accept a shared &Provider or cache in ViewContext, add a retry wrapper around eth_getLogs.

🟠 **MEDIUM** list_versions fails entirely if any version event is missing — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs)
> binary_uri_from_event returns Err when no BinaryVersionPublished log is found. Because list_versions calls it for every version, a single pruned log or provider history cap causes the whole table/status command to fail. Consider making binary_uri optional or skipping individual unresolvable versions instead of aborting the whole listing.

🟠 **MEDIUM** Missing contract address filter in manager's eth_getLogs query — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs)
> The manager's binary_uri_from_event constructs a Filter without .address(self.tangle_address), querying ALL contracts on-chain for BinaryVersionPublished events matching the indexed blueprintId+versionId. The CLI's identical function in cli/src/command/upgrade.rs:794 includes .address(ctx.tangle_contract). Because binaryUri is consumed as a download URL (swap.rs::download_and_verify), a matching event from any other contract would cause the manager to download a binary from the wrong source. Fix: add .address(self.tangle_address) before .from_block().

🟠 **MEDIUM** No test coverage for event-based URI resolution or ABI struct change — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs)
> The new binary_uri_from_event helper and the changed get_binary_version/effective_binary_version paths have no unit or integration tests. This is a critical path: a mistake in topic encoding, event decoding, or windowing would break all binary upgrades. There should be at least a mocked-provider test verifying correct topic construction, window traversal, and URI extraction, plus a round-trip test that the BinaryVersion ABI shape decodes the tnt-core 0.19 return data.

🟡 **LOW** Unused stale opentelemetry_api workspace dependency — [`Cargo.toml`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.toml#L397)
> [Line 397](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/Cargo.toml#L397) declares `opentelemetry_api = { version = "0.20", default-features = false }`, but no crate in the workspace references it and it does not appear in Cargo.lock. It is inconsistent with the rest of the OpenTelemetry ecosystem that was just unified on 0.32 (the API is now part of the main `opentelemetry` crate). Remove this dead dependency to avoid confusion and stale-version drift.

🟡 **LOW** Committed build artifact diverges from serde_json::json! key ordering — [`blueprint_apikey.json`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/blueprint_apikey.json)
> The reordered keys no longer match the insertion order of serde_json::json! in build.rs (which produces name→description→version→…). Future build regeneration will write keys in a different order than what's committed, creating noisy diffs against the checked-in version. No functional impact — JSON objects are unordered and no code parses this file as input. Consider either .gitignore-ing the generated artifact or running the build before committing to keep it in sync with the build.rs order.

🟡 **LOW** Generated blueprint metadata ordering conflicts with its generator — [`blueprint_apikey.json`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/blueprint_apikey.json)
> The committed file is now alphabetically sorted, but examples/apikey-blueprint/apikey-blueprint-bin/build.rs writes it with serde_json::to_string_pretty using the original order (name, description, version, master_revision, manager, jobs). Running the apikey-blueprint-bin build will regenerate blueprint_apikey.json and produce an uncommitted diff. Either update build.rs to emit keys in the same alphabetical order, or keep the committed file in the generator's natural order to avoid churn.

🟡 **LOW** Committed JSON field order mismatches build.rs output — [`blueprint_oauth.json`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/blueprint_oauth.json)
> build.rs generates root-level keys in order: name→description→version→master_revision→manager→jobs (serde_json::json! insertion order at examples/oauth-blueprint/oauth-blueprint-bin/build.rs:6-30). The committed file reorders them to: description→jobs→manager→master_revision→name→version. Since build.rs writes directly to root.join("blueprint_oauth.json"), any cargo build will regenerate the file in the old order, making git status show a dirty working tree. Semantic content is identical — this is a cosmetic workflow nit only.

🟡 **LOW** Committed manifest key order diverges from build.rs generator output — [`blueprint_oauth.json`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/blueprint_oauth.json)
> build.rs (examples/oauth-blueprint/oauth-blueprint-bin/build.rs:6-30) constructs the JSON with serde_json::json! in the order name, description, version, master_revision, manager, jobs. serde_json::to_string_pretty preserves insertion order, so a rebuild would write that order. The committed blueprint_oauth.json instead uses alphabetical order (description, jobs, manager, master_revision, name, version). Rebuilding the oauth-blueprint example will rewrite the file and reintroduce a working-tree diff. Fix by regenerating the file from build.rs or, if the sorted form is intended, updating build.rs to emit the same order.

🟡 **LOW** Event signature hash duplicated between publish_version and binary_uri_from_event — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L460)
> publish_version computes topic0 via keccak256("BinaryVersionPublished(uint64,uint64,bytes32,string)") at [line 460](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L460) while binary_uri_from_event uses BinaryVersionPublished::SIGNATURE_HASH (SolEvent trait) at [line 795](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs#L795). These should produce the same hash but maintaining duplicate signature strings creates a maintenance risk — a mismatch would silently break event lookup. Use SIGNATURE_HASH in both places.

🟡 **LOW** No tests for event-scanning URI resolution — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs)
> The new binary_uri_from_event and VersionView integration have no unit or mocked-RPC tests. cli/src/tests/upgrade.rs only covers argument parsing; the gated e2e test explicitly tolerates the facet not being deployed. Add a test with a stub provider/logs that exercises the filter construction, topic encoding, window scanning, and decode path.

🟡 **LOW** Silent decode failures hide malformed logs — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs)
> Logs matching the topic filter but failing BinaryVersionPublished::decode_log are silently skipped, then the function reports 'no BinaryVersionPublished event'. This masks corrupted/ unexpected log shapes. Collect or log decode errors so operators can tell the difference between 'event never existed' and 'event exists but cannot be parsed'.

🟡 **LOW** list_versions assumes contiguous version IDs starting at 0 — [`cli/src/command/upgrade.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/cli/src/command/upgrade.rs)
> Loop for i in 0..count queries getBinaryVersion(blueprint_id, i). If the on-chain versionId counter started at 1 or had gaps, wrong data or errors result. Currently works because the contract uses a 0-starting monotonic counter, but the assumption is implicit. Pre-existing, not introduced by this PR.

🟡 **LOW** Duplicate binary_uri_from_event between manager and CLI with subtle divergence — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs)
> The same function exists in two locations (manager src/upgrade/chain.rs and cli src/command/upgrade.rs) with slightly different implementations — the CLI adds .address() and connects a bare Provider, the manager uses TangleClient's provider wrapper. The TODO in abi.rs says this entire module is temporary pending tnt-core-bindings, but until then, the duplicate increases risk of drift (the address-filter gap is evidence). Consider extracting to a shared helper or at minimum aligning the filter construction.

🟡 **LOW** Log-scanning logic is duplicated with cli/src/command/upgrade.rs — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs#L135-L190)
> binary_uri_from_event ([lines 135-190](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs#L135-L190)) duplicates the same windowed eth_getLogs scan already present in cli/src/command/upgrade.rs. Keeping two copies risks divergent behavior (as already happened with the missing address filter). Consider moving the helper to a shared upgrade-utils module or crate used by both the manager and cargo-tangle.

🟡 **LOW** No unit test coverage for binary_uri_from_event or u64_topic — [`crates/manager/src/upgrade/chain.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/manager/src/upgrade/chain.rs)
> The 117-line binary_uri_from_event (block-scanning loop, topic0+1+2 filter, decode_log with fallback, window boundary arithmetic) and the u64_topic helper have zero unit tests. The CLI e2e test for upgrade (cli/src/tests/upgrade.rs) explicitly documents that the BlueprintsBinaryVersions facet is not deployed on testnet, so integration coverage is also missing. A unit test mocking eth_getLogs would validate: u64_topic encoding, window-boundary arithmetic (especially saturating_sub wrapping), decode_log skip-on-failure, and the no-event-found error path.

🟡 **LOW** No test coverage for unsubscribe return semantics — [`crates/networking/src/blueprint_protocol/behaviour.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/networking/src/blueprint_protocol/behaviour.rs)
> The method now returns `bool` from gossipsub and the panic-inducing `.expect(...)` is removed, but there are no unit or integration tests exercising the return value (true when subscribed, false when not) or the no-panic path. Concrete fix: add a test that subscribes, unsubscribes, asserts `unsubscribe` returns `true`, and asserts a second unsubscribe returns `false` without panicking.

🟡 **LOW** unsubscribe return value discarded by caller, was previously a panic source — [`crates/networking/src/blueprint_protocol/behaviour.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/networking/src/blueprint_protocol/behaviour.rs)
> The old code called `.expect("failed to unsubscribe from topic")` on the bool returned by gossipsub's unsubscribe. Since `unsubscribe` returns `true` only if the topic was previously subscribed, this would panic on any normal cleanup path when a node unsubscribes from a topic it wasn't subscribed to. The fix removes the panic, which is correct — the caller at service.rs:793 already ignores the return value. No behavioral regression: the caller never checked the return value before or after.

🟡 **LOW** Relay denied match arm needs `..` for libp2p 0.56 compat — [`crates/networking/src/discovery/behaviour.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/networking/src/discovery/behaviour.rs)
> libp2p 0.56.0 relay::Event::ReservationReqDenied added extra fields beyond src_peer_id (likely reservation_id). The `..` rest pattern is required to destructure only the fields used. Without this change the code would not compile against libp2p 0.56.0. The change is purely a mechanical compatibility fix with zero logic change.

🟡 **LOW** Relay reservation-denial status is ignored — [`crates/networking/src/discovery/behaviour.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/networking/src/discovery/behaviour.rs)
> The match arm `relay::Event::ReservationReqDenied { src_peer_id, .. }` uses `..` to discard the new `status: StatusCode` field added in libp2p 0.56.0. Only the peer id is logged, so the reason for the denial is lost. Concrete fix: bind and log the status, e.g. `debug!(%src_peer_id, ?status, "Reservation request was denied");`.

🟡 **LOW** exporter.rs and server.rs: no inline unit tests — [`crates/qos/src/metrics/opentelemetry/exporter.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/metrics/opentelemetry/exporter.rs)
> The MetricReader impl for ArcPrometheusReader and the metrics_handler match block in server.rs have no direct unit tests. Both are exercised indirectly through integration tests (qos_metrics_demo_test, otlp_export_test), but a simple compile-time guard test — e.g., asserting ArcPrometheusReader implements MetricReader — would catch regressions from future OTEL SDK bumps more directly.

🟡 **LOW** prometheus/server.rs: UNTYPED arm is a silent no-op — [`crates/qos/src/metrics/prometheus/server.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/metrics/prometheus/server.rs)
> The UNTYPED match arm logs nothing. While untriaged metric families are rare, a user-defined metric mistakenly typed UNTYPED would be invisible in the info-level dump. Consider emitting a debug-level log noting the metric family name and type for observability.

🟡 **LOW** remote.rs: gRPC connect error chain drops cause detail — [`crates/qos/src/remote.rs`](https://github.com/tangle-network/blueprint/blob/35d342d02a067db9c6e33a39be6598aa44c4d1dc/crates/qos/src/remote.rs)
> The new tonic::transport::Endpoint::from_shared().map_err(|e| Err(e)) pattern discards chrono/timestamp-based transport diagnostics from Endpoint::connect when the URL is valid but unreachable. The outer match falls through to a generic warn! that logs the raw tonic error string. Consider extracting the connect error from inside the Ok(ep) branch to distinguish invalid-URL from unreachable-endpoint in the operator log.

---
<sub>tangletools · 2026-07-07T23:14:18Z · [trace](https://gist.github.com/tangletools/7c9b87d8beef81097b1479669059afcb)</sub>

tangletools (changes requested):
## ❌ 4 Blocking Findings — `35d342d0`

> Full multi-shot audit completed 8/8 planned shots over 20 changed files. Global verifier still owns final merge decision. | Full multi-shot audit completed 8/8 planned shots over 20 changed files. Global verifier still owns final merge decision.

Full immutable report for this review: [trace](https://gist.github.com/tangletools/7c9b87d8beef81097b1479669059afcb)

Summary comment for this run: [full summary](https://github.com/tangle-network/blueprint/pull/1467#issuecomment-4909868590)

---
<sub>tangletools · 2026-07-07T23:14:18Z · immutable [trace](https://gist.github.com/tangletools/7c9b87d8beef81097b1479669059afcb)</sub>

drewstone:
Addressed the blockers in the latest push:

- **HIGH (security)** — scoped `binary_uri_from_event`'s getLogs filter to `.address(self.tangle_address)` (chain.rs:159); an attacker could otherwise emit a matching `BinaryVersionPublished` from any contract and spoof the binary-download URI. Fixed the misleading comments too.
- **HIGH** — crossbeam-epoch 0.9.18 → 0.9.20 (RUSTSEC-2026-0204); Cargo.lock churn minimized to just that + tnt-core-bindings 0.19.0.
- **MED** — version listing now tolerates a missing publish log (blanks the row) instead of aborting.

**On the CRITICAL prometheus finding — respectfully, false positive for this workspace.** The `protobuf` feature is not enabled, so `Metric::get_counter()` returns `&Counter` (plain_model), whose `get_value() -> f64` is correct; the suggested `.value()` does not exist on that type. Ground-truth: `cargo build -p blueprint-qos` compiles clean as-is, and `.value()` fails to compile. Left unchanged.

**hickory-proto 0.25.2** is forced by libp2p 0.56 (libp2p-dns 0.44) from a pre-existing branch commit, not the tnt-core bump — no 0.24-compatible resolution exists while libp2p 0.56 is required.

tangletools (dismissed):
## ✅ Auto-approved drewstone PR — `be2542df`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-08T01:55:13Z</sub>

tangletools (commented):

| | |
|---|---|
| **Concerns** | 2 (1 low, 1 weak-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 398.2s (2 bridge agents) |
| **Total** | 398.2s |

### 💰 Value — error

value agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

Correctly adapts to tnt-core 0.19 dropping binaryUri from BinaryVersion view by sourcing URI from BinaryVersionPublished event logs; all callers are properly updated and failure paths handled.

- **Integration:** Fully wired. CLI: main.rs calls get_effective_version/list_versions/get_version at lines 2234, 2259, 2266, 2488, 2507 — all updated to pass blueprint_id for log resolution and consume VersionView. Manager: watcher.rs:200 calls effective_binary_version(service_id, blueprint_id), event_handler.rs:935 does the same, rpc.rs:268+357 call get_binary_version — all production paths. No dead surface.
- **Fit with existing patterns:** Fits established patterns: inlined sol! ABI stubs (documented as temporary until tnt-core-bindings ships), receipt-log topic decoding (mirrors publish_version at upgrade.rs:458-473 and attest_version at upgrade.rs:593-610), and separate CLI/manager domain types (VersionView vs BinaryVersionInfo) serving their respective surfaces. No competing approach exists — the view no longer provides the URI b
- **Real-world viability:** Holds up under realistic use: 10K-block bounded windows avoid RPC span caps, newest-first scanning stops at first match, address-scoped filter prevents event spoofing, decode failures silently skipped, list_versions survives pruned events (empty URI), NoVersionsPublished handled at all call sites, topic positions correct (topic1=blueprintId, topic2=versionId). One minor concern: CLI creates a fres
- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge warning:** opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content

### 🔎 Heuristic Signals

🟡 **Cruft: todo added** `crates/manager/src/upgrade/abi.rs`
> +// TODO: replace with tnt-core-bindings typed `IBlueprintBinaryVersions` once published.

### 🎯 Usefulness Audit

🟡 **binary_uri_from_event duplicated across CLI and manager** `[ergonomics]` ``
> cli/src/command/upgrade.rs:776-842 and crates/manager/src/upgrade/chain.rs:139-198 implement the same bounded-window newest-first event-log scanning algorithm (~95% identical). For now this is justified — the CLI crate cannot access the manager's crate-private ChainView method, and each uses a different provider type (standalone ProviderBuilder vs TangleClient). If a shared utility crate emerges that both can depend on, unify them there.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260708T020352Z</sub>

tangletools (dismissed):
## ✅ Auto-approved drewstone PR — `dcabfeac`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-08T02:07:40Z</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `2294ca89`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-07-08T02:16:42Z</sub>

tangletools (commented):
## ⚪ Value Audit — audit-incomplete

| | |
|---|---|
| **Verdict** | audit-incomplete |
| **Concerns** | 1 (1 low) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 14.5s (2 bridge agents) |
| **Total** | 14.5s |

### 💰 Value — error

value agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

### 🎯 Usefulness — error

usefulness agent produced no parseable value-audit JSON.

- **Model:** opencode/deepseek/deepseek-v4-pro
- **Bridge attempts:** 3
- **Bridge error:** opencode/zai-coding-plan/glm-5.2: bridge stream ended without value-audit content; opencode/kimi-for-coding/k2p7: bridge stream ended without value-audit content; opencode/deepseek/deepseek-v4-pro: bridge stream ended without value-audit content

### 🔎 Heuristic Signals

🟡 **Cruft: todo added** `crates/manager/src/upgrade/abi.rs`
> +// TODO: replace with tnt-core-bindings typed `IBlueprintBinaryVersions` once published.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260708T021842Z</sub>

PR #1497, opened 2026-08-14
fix(tangle): fail closed on aggregation policy errors

drewstone:
## Summary

Reset the aggregating consumer after every completed submission future.
Propagate service-manager aggregation policy errors instead of selecting an unsafe fallback.

## Change Class

Selected class: Class D

This change modifies protocol result routing and aggregation policy handling.

## Behavior Contract

The consumer returns the original submission error to its sink caller.
After an error, the consumer returns to its waiting state and can process the next job.
When a blueprint manager exists, aggregation and threshold query failures remain errors.
No failed policy query may silently select direct submission or a guessed threshold.
Failure mode choice (`fail-closed`): policy uncertainty must stop submission.

## Risk And Scope

Security impact: prevents a direct result from bypassing a required aggregated result path.
Compatibility impact: successful queries and the public client API remain unchanged.
Migration notes: none; this is a runtime bug fix.
Rollback plan: revert commit `afe04160` if a downstream policy depends on the old fallback.

## Verification

- `cargo test -p blueprint-tangle-extra --lib aggregating_consumer::tests::failed_submission_resets_state_for_next_job -- --nocapture`
- `cargo test -p blueprint-tangle-extra --features aggregation --lib aggregating_consumer::tests::failed_submission_resets_state_for_next_job -- --nocapture`
- `cargo test -p blueprint-client-tangle --test aggregation_policy -- --nocapture --test-threads=1`
- `cargo test -p blueprint-client-tangle --lib`
- `cargo clippy -p blueprint-client-tangle --tests --all-features -- -D warnings`
- `cargo clippy -p blueprint-tangle-extra --lib --all-features -- -D warnings`

## Harness Evidence

Reproducer added: the sink state test polls a failed future, checks the returned error, and processes a second future.
Negative-path coverage added: the JSON-RPC test returns manager errors for both policy methods.
Key assertions that prove the fix: failed futures reset to `WaitingForResult`; manager errors remain `Error::Contract`.

## Checklist

- [x] Added regression tests for the failed submission state and policy query errors.
- [x] Ran focused unit tests and all-feature clippy checks.
- [x] Preserved the existing public API and error return behavior.

PR #1498, opened 2026-08-14
fix(aggregation): match chain quorum rounding

drewstone:
## Summary

- use ceiling division for count and stake quorum calculations
- add the two-operator 67% regression and update boundary expectations

## Root cause

TNT core uses ceilDiv for aggregation thresholds. The client used integer floor division, so two operators at 6700 bps initialized the aggregation service with one required signer. The chain then rejected the aggregate because it required two.

## Checks

- `cargo test -p blueprint-tangle-extra --features aggregation calculate_required_signers`
- `cargo fmt --all -- --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3087,6 +3087,7 @@ dependencies = [
  "futures-core",
  "hex",
  "k256",
+ "mockito",
  "serde",
  "serde_json",
  "thiserror 2.0.18",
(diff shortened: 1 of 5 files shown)


tangletools (approved):
## ✅ Auto-approved drewstone PR — `82508fcd`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-08-14T09:29:14Z</sub>

drewstone:
Superseded by a clean Class D PR based on current main: the original branch included the already-merged #1497 files.

PR #1499, opened 2026-08-14
fix(aggregation): match chain quorum rounding

drewstone:
## Summary

Use ceiling division for count-based and stake-weighted aggregation thresholds.
Add the two-operator 67 percent regression that exposed the mismatch.

## Change Class

Selected class: Class D

This change modifies the quorum that permits an aggregated result submission.

## Behavior Contract

The client computes the same minimum quorum as TNT core.
A nonzero fractional threshold always rounds up to the next operator or stake unit.
Zero operators retain the existing defensive minimum of one.
Calculated signer counts never exceed the active operator count.
Failure mode choice (`fail-closed`): the client never accepts fewer signers than the contract.

## Risk And Scope

Security impact: prevents an aggregation service from declaring readiness below the contract quorum.
Compatibility impact: fractional thresholds can require one additional signer.
Migration notes: none; this aligns existing client behavior with the contract.
Rollback plan: revert commit `82508fcd` if TNT core changes its threshold formula.

## Verification

- `cargo test -p blueprint-tangle-extra --features aggregation calculate_required_signers`
- `cargo fmt --all -- --check`
- `git diff --check origin/main...HEAD`

## Harness Evidence

Reproducer added: two operators at 6700 basis points must require two signatures.
Boundary coverage includes one, two, three, and large operator sets.
Stake-weighted tests now use the same ceiling rule.

## Checklist

- [x] Added the two-operator regression.
- [x] Matched the TNT core ceiling formula.
- [x] Ran focused tests and formatting.
- [x] Kept the change to one runtime file.

### DIFF
--- a/crates/tangle-extra/src/aggregating_consumer.rs
+++ b/crates/tangle-extra/src/aggregating_consumer.rs
@@ -1269,7 +1269,8 @@ pub mod integration {
             if total == 0 {
                 return 1;
             }
-            let mut required = (total as u64 * threshold_bps as u64) / 10000;
+            let product = total as u64 * threshold_bps as u64;
+            let mut required = product / 10000 + u64::from(product % 10000 != 0);
             if required == 0 {
                 required = 1;
             }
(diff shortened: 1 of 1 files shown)


tangletools (dismissed):
## ✅ Auto-approved drewstone PR — `421ac1a3`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-08-14T09:34:34Z</sub>

tangletools (approved):
## ✅ Auto-approved drewstone PR — `d7a8ecca`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-08-14T10:05:25Z</sub>

PR #1505, opened 2026-08-15
fix(tangle): buffer aggregated result gas

drewstone:
## Summary

This change routes aggregated BLS result submissions through the existing gas fallback helper.

It affects operators that submit aggregated results to the Tangle contract.

The helper applies the signer address during gas estimation, adds the existing 1.3x buffer, and uses the existing 8,000,000 gas floor when estimation underestimates the manager-hook path.

## Change Class

- Class A (docs/tooling only)
- Class B (single-crate behavior)
- Class C (cross-crate/runtime behavior)
- Class D (protocol/security/metadata semantics)
- Selected class: Class D
- Why this class: This changes gas handling for a protocol transaction in `crates/clients/tangle/src/`.

## Behavior Contract

- Current behavior: Aggregated submissions send without an explicit gas limit.
- Intended behavior: Aggregated submissions estimate with the signing address, buffer the estimate, and use the 8,000,000 floor when needed.
- Invariants that must hold: The transaction uses the configured signer, the gas limit is never below the floor, and reverted receipts remain errors.
- Failure mode choice (`fail-closed` or `fail-open`) and rationale: Fail-closed. The shared helper returns transport, pending, and on-chain revert errors.

## Risk and Scope

- Security impact: No authorization or signature rule changes.
- Compatibility impact (APIs, metadata format, configs): No public API or configuration changes.
- Migration notes (if any): None.
- Rollback plan: Revert this commit.

## Verification

- `cargo test -p blueprint-client-tangle gas_fallback_tests --lib` — 6 passed, 0 failed.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.

## Harness Evidence

- Reproducer added/updated: Added `aggregate_result_floor_rejects_underestimated_hook_path` for the observed 1,885,516-gas estimate.
- Negative-path coverage added: The regression test covers the under-estimated gas path and asserts the conservative floor.
- Key assertions that prove the fix: The aggregate method calls `send_transaction_with_fallback_gas` with `SUBMIT_RESULT_MIN_GAS_LIMIT`, and the test proves the floor overrides the under-estimate.

## Checklist

- [x] I followed [`docs/engineering/HARNESS_ENGINEERING_PLAYBOOK.md`](/docs/engineering/HARNESS_ENGINEERING_PLAYBOOK.md).
- [x] I followed [`docs/engineering/HARNESS_ENGINEERING_SPEC.md`](/docs/engineering/HARNESS_ENGINEERING_SPEC.md).
- [x] I used [`docs/engineering/HARNESS_REVIEW_CHECKLIST.md`](/docs/engineering/HARNESS_REVIEW_CHECKLIST.md) while implementing/reviewing.
- [x] I added or updated tests that reproduce the original issue.
- [x] I added or updated negative-path tests for invalid/missing/conflicting input.
- [x] I documented behavior or interface changes in docs/examples when needed.
- [x] I evaluated compatibility/migration impact explicitly.
- [x] I validated local formatting, clippy, and relevant tests.

The autoresearch dependency must pin commit `6302ba849967936b30ed677393664dfe542573d6`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4017,6 +4017,7 @@ dependencies = [
  "ark-bn254 0.6.0",
  "ark-ff 0.6.0",
  "ark-serialize 0.6.0",
+ "axum",
  "blueprint-anvil-testing-utils",
  "blueprint-chain-setup-anvil",
  "blueprint-client-tangle",
--- a/crates/clients/tangle/src/client.rs
+++ b/crates/clients/tangle/src/client.rs
@@ -2984,6 +2984,7 @@ impl TangleClient {
         use alloy_sol_types::SolCall;
 
         let wallet = self.wallet()?;
+        let from_address = wallet.default_signer().address();
         let provider = ProviderBuilder::new()
             .wallet(wallet)
             .connect(self.config.http_rpc_endpoint.as_str())
@@ -3004,15 +3005,13 @@ impl TangleClient {
             .to(self.tangle_address)
             .input(calldata.into());
 
-        let pending_tx = provider
-            .send_transaction(tx_request)
-            .await
-            .map_err(Error::Transport)?;
-
-        let receipt = pending_tx
-            .get_receipt()
-            .await
-            .map_err(Error::PendingTransaction)?;
+        let receipt = send_transaction_with_fallback_gas(
+            &provider,
+            from_address,
+            tx_request,
+            SUBMIT_RESULT_MIN_GAS_LIMIT,
+        )
+        .await?;
 
         Ok(TransactionResult {
             tx_hash: receipt.transaction_hash,
@@ -3356,6 +3355,17 @@ mod gas_fallback_tests {
         assert_eq!(
             buffered_gas_limit(Some(21_000), ERC20_APPROVE_MIN_GAS_LIMIT),
             ERC20_APPROVE_MIN_GAS_LIMIT
+        );
+    }
+
+    #[test]
+    fn aggregate_result_floor_rejects_underestimated_hook_path() {
+        // Tempo's manager hook can make aggregate-result estimation understate the
+        // transaction limit. The shared result floor must win in that case.
+        let underestimated_hook_path = 1_885_516;
+        assert_eq!(
+            buffered_gas_limit(Some(underestimated_hook_path), SUBMIT_RESULT_MIN_GAS_LIMIT),
+            SUBMIT_RESULT_MIN_GAS_LIMIT
         );
     }
 
--- a/crates/tangle-aggregation-svc/src/client.rs
+++ b/crates/tangle-aggregation-svc/src/client.rs
@@ -56,6 +56,18 @@ pub enum ClientError {
     /// Threshold not met
     #[error("Threshold not met: {collected}/{required}")]
     ThresholdNotMet { collected: usize, required: usize },
+    /// Another operator already submitted this task to the chain.
+    #[error("Aggregation task was already submitted to the chain")]
+    AlreadySubmitted,
+}
+
+/// Outcome of waiting for an aggregation task.
+#[derive(Debug)]
+pub enum ThresholdWaitResult {
+    /// The caller can submit this aggregate to the chain.
+    Aggregated(AggregatedResultResponse),
+    /// Another operator already submitted the task.
+    Submitted,
 }
 
 /// HTTP client for the aggregation service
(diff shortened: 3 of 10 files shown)


tangletools (approved):
## ✅ Auto-approved drewstone PR — `6302ba84`

<sub>tangletools · auto-approval · reason: `drewstone_author` · 2026-08-15T11:38:23Z</sub>

tangletools (commented):

| | |
|---|---|
| **Coverage** | 2 of 2 lenses (value, usefulness) |
| **Concerns** | 4 (4 weak-concern) |
| **Heuristic** | 0.0s |
| **Duplication** | 0.0s |
| **Interrogation** | 317.0s (2 bridge agents) |
| **Total** | 317.0s |

### 💰 Value — sound-with-nits

Routes the last unbuffered aggregated-result transaction through the existing gas-fallback helper, and adds service-level 'already submitted' detection so waiting operators exit promptly instead of hanging until timeout; coherent, reuses existing primitives, ships.

- **What it does:** Two coupled behaviors across 5 commits. (1) Gas: `submit_aggregated_result` (crates/clients/tangle/src/client.rs:3008) now uses `send_transaction_with_fallback_gas` with `SUBMIT_RESULT_MIN_GAS_LIMIT` — the same helper + constant already used by ~10 other submit paths in that file (client.rs:918, 1067, 1084, 1422, 1448, 1473, 1688, 2067, 2539, 2557, 2944). (2) Coordination: `get_for_aggregation` re
- **Goals it achieves:** Multi-operator aggregated submission actually terminates cleanly: the operator that loses the submit race learns 'someone else submitted' in one poll interval instead of burning the full `threshold_timeout` and reporting a false ThresholdNotMet/timeout error; and the winning operator's transaction doesn't revert from under-estimated gas (the observed 1.88M estimate vs the manager-hook's real cost,
- **Assessment:** Good change, squarely in the codebase's grain. The gas fix is pure consistency — one straggler call site adopting the established helper. The coordination fix reuses primitives that already existed (`status.submitted` field in types.rs:122, `mark_submitted`, the chain-level `is_job_already_completed` backstop at aggregating_consumer.rs:1177) rather than inventing new machinery; the chain-level rev
- **Better / existing approach:** none — this is the right approach. Searched for alternatives: (a) keeping aggregate bytes readable after submission and relying on the existing `JobAlreadyCompleted` chain revert — worse, it wastes gas on a guaranteed-revert tx and rests on fragile revert-string matching (`is_job_already_completed`, aggregating_consumer.rs:1177-1181); (b) push/notify instead of polling — out of proportion to the H
- **Bridge attempts:** 2
- **Bridge warning:** opencode/kimi-for-coding/k2p7: opencode: opencode error

Closes the last unbuffered tx-send path by applying the established gas-fallback helper to aggregated result submission, and fixes a real stuck-operator failure where waiting operators spun to timeout after a peer's submission was cleaned up.

- **Integration:** Fully reachable on live paths. Gas change: TangleClient::submit_aggregated_result (crates/clients/tangle/src/client.rs:2974) is the sole on-chain send for aggregates, reached via AggregatingConsumer sink -> submit_aggregated_to_chain_with_result (crates/tangle-extra/src/aggregating_consumer.rs:1186) -> AggregatedResult::submit (crates/tangle-extra/src/aggregation.rs:267); also exercised by the anv
- **Fit with existing patterns:** Exactly in the grain. send_transaction_with_fallback_gas already had ~13 call sites in client.rs (lines 918, 1067, 1084, 1422, 1448, 1473, 1688, 2067, 2539, 2557, 2944) and submit_result already used SUBMIT_RESULT_MIN_GAS_LIMIT — this PR just brought the one remaining send (submitAggregatedResult) under the same helper, reusing the same 8M floor and 1.3x buffer. The Submitted-outcome-as-enum (not 
- **Real-world viability:** Handles the multi-operator race three redundant ways: submitted-status polling inside the loop and after timeout (client.rs:341-379; aggregating_consumer.rs:1118-1132), JobAlreadyCompleted matching by name and selector, and a 120s retention window (default, service.rs:65) that exceeds the 60s cleanup interval so the submitted flag stays observable for pollers. Persistence migration is sound: #[ser
- **Bridge attempts:** 2

### 💰 Value Audit

🟡 **Poll loop now exists in two places with the same shape** `[duplication]` ``
> `AggregationServiceClient::wait_for_threshold_or_submitted` (crates/tangle-aggregation-svc/src/client.rs:327-380) and `wait_for_threshold_any_service` (crates/tangle-extra/src/aggregating_consumer.rs:1069-1140) both implement aggregate-then-status-then-sleep polling. The multi-service version cannot reuse the single-client waiter without serializing timeouts across services, and the paired structure (wait_for_threshold vs wait_for_threshold_any_service) predates this change, so this extends an e

🟡 **Same event surfaces as Ok(()) in one facade and Err in the other** `[maintenance]` ``
> When another operator submitted: the consumer path returns Ok(()) (aggregating_consumer.rs:1036-1044) while the strategy path returns Err(StrategyError::AlreadySubmitted) (strategy.rs:480,489) with a doc comment telling callers to treat it as a no-op. The Err is forced by `aggregate`'s return type (AggregatedSignatureResult the caller submits), so it is defensible, but blueprint authors using AggregationStrategy directly must learn one more must-match variant. Weak: acceptable taxonomy cost for 

### 🎯 Usefulness Audit

🟡 **Legacy wait_for_threshold now has zero in-repo callers and converts race-loss into an error** `[ergonomics]` ``
> After strategy.rs switched to wait_for_threshold_or_submitted, wait_for_threshold (crates/tangle-aggregation-svc/src/client.rs:320-334) has no production caller in the repo; it survives as public-API compat and maps Submitted -> Err(ClientError::AlreadySubmitted). External callers that only match ThresholdNotMet will now treat a successful peer submission as a failure. Fine to keep for a published crate, but consider marking it deprecated or #[non_exhaustive]-guarding ClientError so the new vari

🟡 **Submitted-status poll loop duplicated across client and consumer** `[ergonomics]` ``
> The get_aggregated-then-get_status polling with timeout re-check now exists both in AggregationServiceClient::wait_for_threshold_or_submitted (client.rs:341-379) and in wait_for_threshold_any_service (aggregating_consumer.rs:1068-1137), which fans out over N services so it cannot delegate directly. The duplication is pre-existing structure extended consistently, not a new fork; a shared per-service probe helper would keep the two loops from drifting. Does not gate shipping.

---

| Pass | What it asks |
|---|---|

</details>
<sub>value-audit · 20260815T114543Z</sub>

